PR #7, opened 2017-12-14
Rewrite the client

carllerche:
This is a full rewrite of the client. You probably want to review the actual source and not the diff.


src/client/mod.rs
```diff
@@ -1,260 +1,190 @@
...
-        }
-    }
-}
+    pub fn build<T>(&mut self, inner: T) -> Grpc<T>
+    where T: HttpService,
+    {
+        let uri = self.uri.as_mut().expect("service URI is required");
+
+        let scheme = uri.scheme_part()
+            .expect("gRPC service URI must include scheme and authority")
```
seanmonstar: Should we be panicking or returning a `Result`?

Most cases, a user probably constructs a Grpc by hand and so a bad URI is a programmer error, though it could be that it is constructed by config or user input...

examples/helloworld_client.rs
```diff
@@ -0,0 +1,106 @@
...
+    let conn = Conn(addr, reactor.clone());
+    let h2 = tower_h2::Client::new(conn, Default::default(), reactor);
+
+    let done = h2.new_service()
+        .map_err(|e| unimplemented!("h2 new_service error: {:?}", e))
+        .and_then(move |service| {
+            let uri = "http://127.0.0.1:8888/".parse().unwrap();
+            let mut client = Greeter::new(service, uri);
+
+            client.say_hello(grpc::Request::new(HelloRequest {
```
seanmonstar: It'd be nicer to be able to write `client.say_hello(HelloRequest { ... })` instead of wrapping in a `grpc::Request`. Adding metadata doesn't seem super common to need to be able to do it with every rpc.

examples/helloworld_client.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub fn poll_ready(&mut self) -> futures::Poll<(), grpc::Error<T::Error>> {
+        self.inner.poll_ready()
+    }
+
+    pub fn say_hello(&mut self, request: grpc::Request<HelloRequest>)
+        -> grpc::unary::ResponseFuture<HelloReply, T::Future, T::ResponseBody>
+    where grpc::unary::Once<HelloRequest>: Encodable<T::RequestBody>,
+    {
+        let path = http::uri::PathAndQuery::from_static("/helloworld.Greeter/SayHello");
+        self.inner.unary(request, path)
```
seanmonstar: Why isn't the path part of the `grpc::Request`? It seems more appropriate there...

carllerche: I removed it because that information is never supplied by the user (it is set by the client and server lib) and the user can get that info by knowing which grpc fn they are calling / being called from. 

Even if the user could read the value, they don’t know what the value should be when creating the grpc request struct which means it would need to be set to None (using a default value would mean if the request were passed on to a different fn. and that Fn read the default value, they wouldn’t have anyway to know it was a default value and not a real value)

Removing it also means the struct size can be reduced.

Overall it made things much cleaner.

examples/helloworld_client.rs
carllerche: It seemed better to me to be explicit here vs figuring out an advanced api vs simple api.

I considered an Into strategy but for this initial rewrite it seemed out of scope. I think if we want to pursue this discussion we should create a separate issue to discuss and assign that issue to the 0.1 milestone.

I would be happy to have an eronomic api but I think one has to be able to set the metadata to.

src/client/mod.rs
carllerche: A result here is probably fine.

examples/helloworld_client.rs
seanmonstar: I agree that one should have a way to set the metadata, but I think it is less common. Additionally, there seems to already exist a way to do so, using the inner `Grpc::{unary, etc}` methods.

Anywho, I'm fine with making things easier in a future issue.

examples/helloworld_client.rs
seanmonstar: I'm not sure I agree, it's the same as someone needing to know the `Uri` when they want to create an `http::Request`. As for the size of the struct, the `Uri` needs to be passed as a required separate argument, so I don't see a gain.

carllerche: I don't see how it is the same as with HTTP. In the HTTP case you *have* to set the URI to send a request. An HTTP client has no way of knowing what the URI should be.

In the case of the gRPC client (at least the protobuf one), you would never set the URI, the URI is going to be set internally to the client. In fact, the client is not expected to even know what the request URI should be as it is an internal detail of codegen. All the issues that I outlined above are not true for an HTTP client.

Either way, not exposing a URI is the more conservative option here, if I am wrong, we can always add it back later.

PR #10, opened 2017-12-26
Preliminary support for .proto files with multiple services

### LINKED ISSUE #9: Codegen generates invalid code for proto files containing multiple service definitions
Noticed this while working on #3 (gRPC interop test cases). When a `.proto` file defines multiple services, `tower-grpc-build` will generate a separate `pub mod server` and `pub mod client` for each service in the generated code's root module. Needless to say, the Rust compiler doesn't like this.

Failing example (I modified `helloworld.proto`):
```protobuf
syntax = "proto3";

package helloworld;

service Greeter {
  rpc SayHello (HelloRequest) returns (HelloReply) {}
}

service Welcomer {
  rpc SayHi (HelloRequest) returns (HelloReply) {}
}

message HelloRequest {
  string name = 1;
}

message HelloReply {
  string message = 1;
}
```
and compiler errors:
```rust
   Compiling tower-grpc-examples v0.1.0 (file:///Users/eliza/Code/tower-grpc/tower-grpc-examples)
error[E0428]: the name `client` is defined multiple times
   --> /Users/eliza/Code/tower-grpc/target/debug/build/tower-grpc-examples-e932e45db921086e/out/helloworld.rs:232:1
    |
14  | / pub mod client {
15  | |     use ::tower_grpc::codegen::client::*;
16  | |     use super::{HelloRequest, HelloReply};
17  | |
...   |
44  | |     }
45  | | }
    | |_- previous definition of the module `client` here
...
232 | / pub mod client {
233 | |     use ::tower_grpc::codegen::client::*;
234 | |     use super::{HelloRequest, HelloReply};
235 | |
...   |
262 | |     }
263 | | }
    | |_^ `client` redefined here
    |
    = note: `client` must be defined only once in the type namespace of this module

error[E0428]: the name `server` is defined multiple times
   --> /Users/eliza/Code/tower-grpc/target/debug/build/tower-grpc-examples-e932e45db921086e/out/helloworld.rs:265:1
    |
47  | / pub mod server {
48  | |     use ::tower_grpc::codegen::server::*;
49  | |     use super::{HelloRequest, HelloReply};
50  | |
...   |
229 | |     }
230 | | }
    | |_- previous definition of the module `server` here
...
265 | / pub mod server {
266 | |     use ::tower_grpc::codegen::server::*;
267 | |     use super::{HelloRequest, HelloReply};
268 | |
...   |
447 | |     }
448 | | }
    | |_^ `server` redefined here
    |
    = note: `server` must be defined only once in the type namespace of this module

error: aborting due to 2 previous errors
```

hawkw:
This pull request adds support for .proto files which define multiple services to `tower-grpc-build`. 

### Changes

Previously (see #9), when encountering such a file, `tower-grpc-build` would generate separate `client` and `server` modules for each service in the root namespace of the Rust module generated for that protobuf file. Since module names must be unique within their namespace, these generated files would fail to compile. I've modified `tower-grpc-build` to output the code generated for each service inside a single `client` or `server` module.

Unfortunately, although it seems to work correctly, the resulting implementation of this behaviour is a pretty ugly hack. This is due in part to how `prost` works and to the constraint that the code generated for each service be output in a single `client` or `server` module (rather than namespaced in a separate module for each service). 

If you're the sort of person who can't look away from a train wreck or automobile accident, read on to learn about this extremely bad code...

### Here's how it works

In order to make this change, I made an upstream change in the `codegen` crate (PR carllerche/codegen#1) to allow modules in a `codegen::Scope` to be accessed by name once they have been created, and to get a module if it exists or create it if it does not. I then changed the way service generation works so that, rather than constructing a new `Scope` in each call to `{client, server}::ServiceGenerator::generate`, a single `Scope` representing the root scope of the file is passed between each call to `generate`, and the code generated for each service is _added_ to the existing `client` or `server` modules.

Here's where it gets tricky. Since `prost-build` is not aware of the `codegen` crate, it expects each `ServiceGenerator::generate` invocation to write out the generated code to the buffer that will eventually be written out to the file. However, because we're changing existing modules, we can't simply append new code to the end of the buffer anymore; we now have to re-generate the whole `Scope`. This means that multiple `ServiceGenerator::generate` invocations would now append the current state of the `codegen::Scope` – i.e., all the generated code, including the code generated for previous services – to the buffer. 

Ideally, we'd be able to `fmt` the scope a single time, after code for all services has been generated, but we can't, because at this point we've already ceded control to `prost_build`; after
https://github.com/tower-rs/tower-grpc/blob/99c433a7369c18a5d3edf1a0ba5ab726ece9f747/tower-grpc-build/src/lib.rs#L89-L93
rather than calling into `prost`, `prost` is now calling _us_.  Furthermore, any given invocation of `generate` has no way of knowing if it's the _last_ service to be generated, so we can't just wait to `fmt` the scope until the last service is generated.

Initially, I tried having every call to `generate` clear the passed-in buffer before writing the generated code back into it; so that only the code written to the buffer by the final service to be generated would actually end up in the file. This is ugly enough as it is, but it also didn't work, because it also wiped out all the code generated by `prost`. This code doesn't exist in our `codegen::Scope` and was written to the buffer before `prost_build` started calling `ServiceGenerator::generate` for each service.  

What I ended up having to do is track the number of characters written to the buffer by the last 
invocation of `ServiceGenerator::generate`, 
https://github.com/tower-rs/tower-grpc/blob/99c433a7369c18a5d3edf1a0ba5ab726ece9f747/tower-grpc-build/src/lib.rs#L51-L55
and only drop this much text from the buffer before generating each service.
https://github.com/tower-rs/tower-grpc/blob/a844add77419213ad9d582055fb4718a33feabae/tower-grpc-build/src/lib.rs#L107-L110
 This way, the code generated by `prost` should be left intact, but only the final iteration of the `Scope` containing our generated code makes it into the file. Of course, this approach is pretty inefficient, adds additional complexity, and is just plain ugly; I'd like to find a better way to handle this, but at least it _works_ for now.

Fixes #9


hawkw:
This branch was accidentally based on my development branch `eliza/interop-test` rather than on `master` --- sorry about that, my bad.

vorner:
I can confirm it works for me (but I haven't looked at the code).

However, ideas that come into mind for a nicer solution:

* Swap the namespaces, eg instead of `client::ServiceName`, create `service_name::Client`. But I guess this is not really wanted.
* Make an upstream change to prost-build. The `ServiceGenerator` trait could get another method `fn finalize(&self, buf: &mut String);`. The whole thing could be output in this method. If the method had a default (empty) implementation, it would be backwards-compatible change.

hawkw:
@vorner 
> Swap the namespaces, eg instead of `client::ServiceName`, create `service_name::Client`. But I guess this is not really wanted.

Yeah, that was the first thought I had, too; after discussing with @carllerche I believe that structuring the generated code this way is preferred.

> Make an upstream change to prost-build. 

Yeah, some kind of change to `prost-build` is the eventual goal; this is a stop-gap solution until changes can be made upstream.

vorner:
I'm a bit familiar with `prost`s code and this seemed like a really simple change, so I sent a pull request there.

hawkw:
Thanks, @vorner!

carllerche:
It looks like the PR was merged!

hawkw:
Oh, excellent! I'll look at rewriting this PR with the `prost` changes, then.

carllerche:
The addition hasn't made it to crates.io, but hopefully it will soon.

vorner:
A release is said to be planned, but it waits for this to get resolved first: https://github.com/rust-lang/rust/issues/47175.

hawkw:
@carllerche I just cleaned up the git history for this branch, WDYT?

PR #11, opened 2017-12-27
Set content type of request with associated type

nak3:
Even though Codec traits has CONTENT_TYPE as associated constants,
current code hard codes it.

This patch changes to add content type by the associated content type
of the implementation of Codec.


carllerche:
Thanks for the PR.

Removing the hard coded content type would be nice. However, I believe that the two changes are in different code paths. One is the client (the removal) and the other is the server (the addition). The request received by the server should already have the content type set.

Unfortunately we don't have any tests yet (we are working on getting the interopt tests going)...

PR #13, opened 2017-12-29
Encode only generates trailers on the server side

### LINKED ISSUE #12: RPC from tower-grpc client to grpc-go server fails with broken pipe
When a `tower-grpc` client calls a service on a server using the `grpc-go` gRPC implementation, the RPC fails with a broken pipe error:
```
Inner(Error { kind: Inner(Error { kind: Io(Error { repr: Kind(BrokenPipe) }) }) })
```

### Background

I first encountered this issue working on #3 (implementing the gRPC interop tests). I was working on writing the client for the interop tests, and as I had not yet written a server implementing the test RPCs, I was testing my client against the `grpc-go` interop test server, which seemed like a canonical reference implementation. Attempting to call any of the RPCs in the interop test proto resulted in this error. 

I initially suspected a bug in my code, but after trying an [extremely stripped-down client](https://github.com/tower-rs/tower-grpc/blob/cd0d7ed3ceb607d3c445b85a5ef535c9e9cc6995/tower-grpc-examples/src/interop/empty_unary_client.rs) copied from the hello world client in `tower-grpc-examples` that only calls the simplest RPC on the test service, an empty unary call, I wondered if it was really my fault. I then tested the hello world client from the latest `tower-grpc` master against the hello world server in `grpc-go`. Since these example programs are extremely simple and don't rely on any features that `tower-grpc` doesn't currently implement, I'd expect them to be interoperable. However, trying the `tower-grpc` `helloworld-client` against `grpc-go`'s `greeter_server` fails with the same broken pipe error. When the `tower-grpc` client calls `tower-grpc`'s `helloworld-server`, or a Go client calls the Go server, everything works fine.

### What seems to be going wrong

I tried using `tshark` to sniff the connection between the client and server, and it appears that what's happening is that after the server receives the request, it responds with a TCP `ACK` packet followed by a `FIN, ACK` packet. The client complies with the `FIN, ACK` and closes the connection. I have no idea _why_ `grpc-go` is flat-out refusing to talk to us, though.

Here's [the complete output of the `tshark` capture](https://gist.github.com/hawkw/1c5818f17962c4aa2e531ef5f2896f84#file-tshark-capture), and here's the [logs output from the client with `RUST_LOG=trace`](https://gist.github.com/hawkw/1c5818f17962c4aa2e531ef5f2896f84#file-helloworld-log). 

I'd like to get logs out of the Go server as well, but I haven't been able to figure out _how_ yet; it doesn't seem to be documented anywhere as far as I can tell. Maybe a Go person could give some pointers in the right direction.

### Steps to reproduce
(assuming a working Rust and Go installation)
1. check out the latest master of `tower-grpc`.
2. get `grpc-go`, including dependencies for examples, with:
```
$ go get -u google.golang.org/grpc/...
```
3. start the hello world example server from `grpc-go`:
```
$ cd $GOPATH/src/google.golang.org/grpc/examples/helloworld
$ go run greeter_server/main.go
```
4. from the `tower-grpc` repo, run the hello world client:
```
$ cargo run -p tower-grpc-examples --bin helloworld-client
```
And that's it – the client should fail with a `BrokenPipe` error.

hawkw:
Fixes #12.

In 16c02a216a3f45779878c95dc492ae5e52bb0665, the client and server were refactored to share one generic `Encode` implementation. However, this resulted in client request streams ending in trailers when they weren't supposed to, breaking compatibility with other gRPC implementations (see https://github.com/tower-rs/tower-grpc/issues/12#issuecomment-354369747).

I added a boolean `return_trailers` field on `Encode`, which is now set `true` by servers and `false` by clients. 

Tested against `grpc-go` and `grpc-java`'s `helloworld` examples, and the issue described in #12 no longer occurs.

### DIFF
--- a/src/generic/codec.rs
+++ b/src/generic/codec.rs
@@ -143,11 +143,11 @@ where T: Encoder<Item = U::Item>,
         }
     }
 
-    pub(crate) fn error(status: Status, return_trailers: bool) -> Self {
+    pub(crate) fn error(status: Status) -> Self {
         Encode {
             inner: EncodeInner::Err(status),
             buf: BytesMut::new(),
-            return_trailers,
+            return_trailers: true,
         }
     }
 }
--- a/src/generic/codec.rs
+++ b/src/generic/codec.rs
@@ -64,6 +64,9 @@ pub struct Encode<T, U> {
 
     /// Destination buffer
     buf: BytesMut,
+
+    /// Set to true when trailers should be generated.
+    return_trailers: bool,
 }
 
 #[derive(Debug)]
@@ -132,17 +135,19 @@ impl<T, U> Encode<T, U>
 where T: Encoder<Item = U::Item>,
       U: Stream,
 {
-    pub(crate) fn new(encoder: T, inner: U) -> Self {
+    pub(crate) fn new(encoder: T, inner: U, return_trailers: bool) -> Self {
         Encode {
             inner: EncodeInner::Ok { encoder, inner },
             buf: BytesMut::new(),
+            return_trailers,
         }
     }
 
-    pub(crate) fn error(status: Status) -> Self {
+    pub(crate) fn error(status: Status, return_trailers: bool) -> Self {
         Encode {
             inner: EncodeInner::Err(status),
             buf: BytesMut::new(),
+            return_trailers,
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/generic/codec.rs
```diff
@@ -132,17 +135,19 @@ impl<T, U> Encode<T, U>
...
+    pub(crate) fn new(encoder: T, inner: U, return_trailers: bool) -> Self {
         Encode {
             inner: EncodeInner::Ok { encoder, inner },
             buf: BytesMut::new(),
+            return_trailers,
         }
     }
 
-    pub(crate) fn error(status: Status) -> Self {
+    pub(crate) fn error(status: Status, return_trailers: bool) -> Self {
```
seanmonstar: Considering that the only way to send an error is via trailers, this can probably leave the argument off and just always be true.

hawkw: Good point.

PR #15, opened 2018-01-02
Add functions for accessing Request header map

hawkw:
I've added accessor functions to `Request` for getting the header map. I need to add headers to a request for [one of the gRPC interop test cases](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#cacheable_unary), and this seemed like the most ergonomic way to do it.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -25,6 +25,16 @@ impl<T> Request<T> {
         &mut self.message
     }
 
+    /// Get a reference to the request headers.
+    pub fn headers(&self) -> &http::HeaderMap {
+        &self.headers
+    }
+
+    /// Get a mutable reference to the request headers.
+    pub fn headers_mut(&mut self) -> &mut http::HeaderMap {
+        &mut self.headers
+    }
+
     /// Consumes `self`, returning the message
     pub fn into_inner(self) -> T {
         self.message


carllerche:
I'm good with this.

Since the change implies that the full set of HTTP headers are exposed as part of the public API., I'd like @sean to weigh in as well.

hawkw:
> Since the change implies that the full set of HTTP headers are exposed as part of the public API

If we think this should be more restrictive, we could only expose the ability to add headers (and not to delete or modify existing headers)? It seems like [this is what `grpc-go` does](https://godoc.org/google.golang.org/grpc#SetHeader).

seanmonstar:
What header is needed to be set? Is it specific to gRPC, and so perhaps should have a dedicated method instead?

hawkw:
@seanmonstar in the case of this specific test, the header is `x-user-ip`, which is not gRPC-specific.

seanmonstar:
Hm, do we really need to be able to set this? It's just for when testing over some proxies, right?

hawkw:
> do we really need to be able to set this?

That's above my pay grade; it was in the test description so I assumed we should do it :)

With that said, since `grpc-go` exposes a `SetHeader` function, I'm guessing there are some other use cases for this; `cacheable_unary` is, however, the only interop test currently that requires setting a header manually.

seanmonstar:
So, my gut says that perhaps these things don't belong in gRPC directly. It already wraps an `HttpService`, so that wrapped service can inject any sort of HTTP non-sense it wants. Perhaps it can be done at that layer?

hawkw:
(I was going to fall back to checking how grpc-go implements this test case, but it turns out it....doesn't.)

carllerche:
@seanmonstar One should be able to set arbitrary headers from the gRPC application. The HTTP headers is the strategy by which user land "metadata" is communicated between processes.

It is possible that this specific case would be better handled pre-gRPC, but that wouldn't be universally true.

The one question in my mind is, should gRPC specific headers be exposes to the end user? Simply exposing the HeaderMap would mean that this is true. That said, I am not super worried about doing so personally.

seanmonstar:
If one has to be able to set arbitrary headers, then that seems in conflict with exposing gRPC specific headers.

I'm not actually worried about it. Perhaps the gRPC client would set relevant headers at the end, after a user has possibly tried to set headers incorrectly?

hawkw:
How do we feel about exposing methods to _set_ headers but not to access or modify existing headers, à la `grpc-go`?

That seems like the easiest solution for now, and if we decide later that we should allow access to headers, we can add `get_header` (or whatever) in a way that's backwards-compatible with an existing `set_header` --- no breaking existing api.

carllerche:
@seanmonstar The gRPC library get the final say with setting headers as is. I'm not super worried about conflict, and as such I'm ok with the PR as is :)

Issue #16, opened 2018-01-03
client does not properly send streaming requests


hawkw:
I've run into some issues that suggest a probable bug in `tower-grpc` while working on implementing the [`client_streaming`](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md#client_streaming) interoperability test case. 

This test has the client call a client streaming RPC, sends multiple payloads, and then await a unary response, while the server aggregates the number of bytes received from the client and sends that value in the response. The client then asserts that the server received the correct number of bytes.

Here's the relevant code in the client implementation:

https://github.com/tower-rs/tower-grpc/blob/6a7ce4077200080cbd47b2bb6e209b087316a653/tower-grpc-examples/src/interop/client.rs#L390-L416

When I run the `tower-grpc` test client against `grpc-go`'s interop server, the test fails at the assertion that the bytes aggregated value returned by the server is correct:
```
client_streaming:
  ✔ call must be successful
  ✖ aggregated payload size must be 74922 bytes
  in `response.aggregated_payload_size == 74922`: aggregated_payload_size=0
```

Interestingly, when I run the same test against `grpc-java`'s test server, the server returns an error, and prints the following stack trace to the console:
```
Jan 03, 2018 12:55:27 PM io.grpc.internal.SerializingExecutor run
SEVERE: Exception while executing runnable io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1MessagesAvailable@3ac811
io.grpc.StatusRuntimeException: INTERNAL: Invalid protobuf byte sequence
	at io.grpc.Status.asRuntimeException(Status.java:517)
	at io.grpc.protobuf.lite.ProtoLiteUtils$2.parse(ProtoLiteUtils.java:168)
	at io.grpc.protobuf.lite.ProtoLiteUtils$2.parse(ProtoLiteUtils.java:82)
	at io.grpc.MethodDescriptor.parseRequest(MethodDescriptor.java:287)
	at io.grpc.internal.ServerCallImpl$ServerStreamListenerImpl.messagesAvailable(ServerCallImpl.java:251)
	at io.grpc.internal.ServerImpl$JumpToApplicationThreadServerStreamListener$1MessagesAvailable.runInContext(ServerImpl.java:624)
	at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37)
	at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:123)
[...]

Issue #14, opened 2017-12-29
Hang with grpc-interop "large_unary" test


hawkw:
A `tower-grpc` client hangs when running the `large_unary` test in the gRPC interoperability test cases against the `grpc-go` test server implementation.

This test involves sending a 271828-byte request and receiving a 314159-byte response. Looking at the logs, I suspect there's a flow control issue (in `h2`?) involved here; it seems like we run out of window capacity and it's never released, I'm not sure why yet.

Client logs: https://gist.github.com/hawkw/e6dda003747fbe87f2ae72d1f096ce0c#file-client-log
Tshark capture: https://gist.github.com/hawkw/e6dda003747fbe87f2ae72d1f096ce0c#file-tshark

After reaching printing the last line in the logs, no additional output was seen from the client for several minutes.

hawkw:
Interesting: running this against the `grpc-java` implementation of the interop test server panics with
```
thread 'main' panicked at 'assertion failed: stream.send_flow.available() > 0', /Users/eliza/.cargo/git/checkouts/h2-5627e19d2c95dbee/fc75311/src/proto/streams/prioritize.rs:373:12
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
but _only if this was the first test run against the server since it was started_; otherwise, it still just hangs. I have [H2 logs](https://gist.github.com/hawkw/e6dda003747fbe87f2ae72d1f096ce0c#file-client-with-panic-log) from that run as well.

seanmonstar:
The logs show that 5 bytes are released back to the window when reading the body, and a gRPC framed message has 5 bytes for the header. However, it doesn't release the rest of the bytes since it doesn't finish decoding the message.

It seems like handling this would require the client to know before hand the max message size that can be received, and set the initial window size accordingly.

hawkw:
Panic no longer occurs after carllerche/h2@26e7a2d4165d500ee8426a10d2919e40663b4160; the client now hangs in all cases.

hawkw:
@seanmonstar as far as I can tell, the gRPC spec doesn't formally specify a max message size, and most implementations have it user configurable. It looks like in most cases, the default is 4 MB.

I believe most gRPC implementations set default window sizes of 64k and then tune them based on BDP. For now, is it correct for us to set an initial window size of 4MB?

seanmonstar:
What I meant by client was the user of the grpc client. When creating a `Client`, they should probably configure the h2 connection first.

[...]

PR #19, opened 2018-01-08
Run interop tests on CI

hawkw:
This PR adds the beginning of an implementation of the gRPC interoperability tests, and a first pass at running them on CI. Although this implementation is far from complete, I want to get the tests that _are_ currently implemented merged in so we can start running them against other changes.

## Current state of test implementation:

We have a test client implementation with a framework for adding more tests. We have not yet implemented a test server, and the `docker-compose` file for running the tests currently runs them against `grpc-go`'s test server for now.

### Test checklist

- [x] `empty_unary`: implemented in client
- [ ] `cacheable_unary`: started, requires request context implementation to set cacheable flag
- [x] `large_unary`: implemented in client, broken due to #14
- [ ] ~`client_compressed_unary`~: requires gRPC compression, NYI
- [ ] ~`server_compressed_unary`~: requires gRPC compression, NYI
- [x] `client_streaming`: implemented in client, broken due to #16
- [ ] ~`client_compressed_streaming`~: requires gRPC compression, NYI
- [ ] `server_streaming`
- [ ] ~`server_compressed_streaming`~: requires gRPC compression, NYI
- [ ] `ping_pong`
- [ ] `empty_stream`
- [ ] ~`compute_engine_creds`~ requires auth, NYI
- [ ] ~`jwt_token_creds`~ requires auth, NYI
- [ ] ~`oauth2_auth_token`~ requires auth, NYI
- [ ] ~`per_rpc_creds`~ requires auth, NYI
- [ ] `custom_metadata`
- [ ] `status_code_and_message`
- [ ] `unimplemented_method`
- [ ] `unimplemented_service`
- [ ] `cancel_after_begin`
- [ ] `cancel_after_first_response`
- [ ] `timeout_on_sleeping_server`
- [ ] ~`concurrent_large_unary`~: not yet added to test case descriptions

I've added a `docker-compose.yml` and two `Dockerfile`s for running the test client against the `grpc-go` test server without requiring a Go installation on the host machine. This is now used in the `.travis.yml` for running the tests, and might be useful in dev as well. I also added a quick README briefly documenting how to run the interop tests.


hawkw:
Note that _only_ the `empty_unary` test case is actually run on Travis currently --- it is the only test case that actually passes. The other test cases are either not yet implemented, rely on functionality that `tower-grpc` doesn't yet implement, or are currently broken, as I noted above.

tower-grpc-build/Cargo.toml
```diff
@@ -4,5 +4,6 @@ version = "0.1.0"
 authors = ["Carl Lerche <me@carllerche.com>"]
 
 [dependencies]
-codegen = { git = "https://github.com/carllerche/codegen" }
+codegen = { git = "https://github.com/hawkw/codegen", branch = "eliza/get-or-create"}
```
olix0r: we shouldn't merge this as-is

tower-grpc-build/src/lib.rs
```diff
@@ -22,10 +22,17 @@ struct Inner {
     build_server: bool,
 }
 
+struct GeneratorState {
+    root: codegen::Scope,
+    // XXX this is where it gets _really_ ugly.
```
olix0r: is this comment actionable? should something change? if not, please remove.

tower-grpc-build/Cargo.toml
hawkw: Carl just merged this branch (carllerche/codegen@533ca264ebd251ea5876828fccf58275e3c91410), so this can now depend on master

tower-grpc-interop/Dockerfile-server-go
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.9
+
+ENV PACKAGE google.golang.org/grpc/interop/server
+ENV PACKAGEDIR $GOPATH/src/$PACKAGE
+
+WORKDIR /
```
olix0r: what is needed from the workdir? As it is, this line bsaically ensures that you don't cache the rest of the layers below this

tower-grpc-build/src/lib.rs
hawkw: some upstream changes were recently made in `prost-build` (danburkert/prost#65) which will let us make that code much less ugly (see discussion on PR #10), so this can hopefully go away soon.

tower-grpc-interop/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: '3'
+services:
+
+  server-go:
+    build:
+      context: .
+      dockerfile: Dockerfile-server-go
```
olix0r: do we really need to rebuild this docker image on every build? can't we just publish it? what does it need from the local repo?

tower-grpc-interop/Dockerfile-server-go
```diff
@@ -0,0 +1,11 @@
...
+RUN go get -u $PACKAGE
+WORKDIR $PACKAGEDIR
+
+RUN ["go", "install", "."]
+ENTRYPOINT ["server"]
```
olix0r: If this docker image can be published (which i think it should be, because i don't think it makes much sense to rebuild on every CI/etc), then we should make this a multistage build so that the resulting image only includes the binary and not the full golang image.

tower-grpc-build/src/lib.rs
olix0r: I'd prefer that we take out anything like "this is ugly" or "this is a terrible hack" and replace it with something actionable.

tower-grpc-interop/docker-compose.yml
hawkw: i wasn't sure whether a docker image of grpc-go's interop test server was something we would _want_ to publish or not? if it is, i can certainly try and make that change.

tower-grpc-interop/Dockerfile-server-go
hawkw: this is just a case of me not knowing a whole lot about Docker; @siggy sent me an example dockerfile for a Go program and this is based directly on that. will fix.

tower-grpc-build/src/lib.rs
hawkw: I've changed this comment in the branch that this branch should logically be based on (#10); when I rebase this branch onto that branch, this comment will be replaced. It now looks like this:
https://github.com/tower-rs/tower-grpc/blob/c7572dd92dda90c5bc2f484c41b323a1557834d4/tower-grpc-build/src/lib.rs#L32-L35

Is that better?

tower-grpc-interop/Dockerfile-server-go
hawkw: Addressed in 933a6875d4fc700ac25e35e834291915c221a5cf.

tower-grpc-interop/Dockerfile-server-go
hawkw: Addressed in 933a6875d4fc700ac25e35e834291915c221a5cf.

hawkw:
@olix0r, commit 2be62e9e11444eac031c04bd26b79ee1c8df7959 adds a quick Bash script for running the interop tests (you were right, that _did_ end up being much easier), and 5c44d82cc52b82a97c984da7450f4575364397f5 changes Travis to use that rather than Docker Compose. I've left the Docker stuff in the repo as I thought it could potentially be useful for running the interop tests locally, but if you'd prefer, I can remove them completely.

tower-grpc-examples/build.rs
```diff
@@ -6,12 +6,19 @@ fn main() {
...
     tower_grpc_build::Config::new()
         .enable_server(true)
         .enable_client(true)
         .build(&["proto/routeguide/route_guide.proto"], &["proto/routeguide"])
-        .unwrap();
-}
+        .unwrap_or_else(|e| panic!("protobuf compilation failed: {}", e));
+
+    // Build grpc-interop
+    tower_grpc_build::Config::new()
```
carllerche: Should this be here or in the `tower-grpc-interop` crate?

I don't really have much of an opinion either way, it was initially unexpected to me.

hawkw: Whoops, I meant to remove that --- I think I forgot to commit it. `tower-grpc-interop` *should* have its own `build.rs` unless I also forgot to commit _that_...

hawkw:
@carllerche this passes ci on stable Rust but fails on nightly/beta; I think that's due to the same issues with `ring` that we've seen elsewhere?

hawkw:
@carllerche 
> Looks good to me. I can't really comment on the docker stuff, if it works, great. 

The Docker stuff is no longer used on CI, as @olix0r suggested that it could be made simpler by replacing it with a shell script. I've left it, in case it's useful for running `grpc-go`'s test server in dev (the CI script only works on linux-amd64).

PR #21, opened 2018-01-09
Rewrite ServiceGenerator implementations to use finalize()

hawkw:
Now that danburkert/prost#63 has landed (thanks @voerner!), `tower-grpc-build` can use `ServiceGenerator::finalize()` for generating multiple services, rather than the workaround added in #10. The resultant code is much cleaner.

Note that this now means we depend on `prost` from Git, rather than `crates.io`, until the code added in danburkert/prost#63 makes it into a published version of `prost`; other changes make recent revisions of `prost-build` incompatible with the currently published `prost-derive` and `prost` crates, so we must get those dependencies from Git as well (20c8546dc25d962f629121ecaddde61c5c2875ea). I have, however, ensured that the `tower-grpc-examples` and `tower-grpc-build` Cargo.tomls depend on the same _revision_ of the `prost `repository. Once `ServiceGenerator::finalize` makes it into a `prost` release, it should be trivial to go back to getting the dependency from crates.io.


tower-grpc-build/src/lib.rs
```diff
@@ -101,46 +84,40 @@ impl Config {
...
-        // `prost` is in there and we need that.
-        // XXX rewrite to use `ServiceGenerator::finalize` from `prost-build`.
-        if state.last_gen_length > 0 {
-            let new_len = buf.len() - state.last_gen_length;
-            buf.truncate(new_len);
-        }
 
-        // Add an extra new line to separate messages
-        write!(buf, "\n").unwrap();
+    fn generate(&self, service: prost_build::Service, _buf: &mut String) {
```
danburkert: `prost` could probably further simplify this by making `ServiceGenerator::generate` and `ServiceGenerator::finalize` take `&mut self`.  Then you wouldn't have to mess with refcell.  @vorner what do you think?

vorner: I thought about that and decided it would be better, but it would also be a backwards-incompatible change, so I stayed with the original shared borrow in the pull request.

carllerche: Switching to `&mut self` would be better for us :)

PR #29, opened 2018-01-10
Immediately release capacity for received data.

carllerche:
Received data is buffered until a full gRPC message is received, at
which point it is decoded. Prior to this patch, the H2 capacity was not
released to the peer until *after* the deserialization happened. This
means that if the H2 window is not large enough, a dead lock is hit.

This patch changes the behavior to immediately release the H2 capacity
when the data is received. This matches how Go's gRPC implementation
works and is also not materially different than the prior behavior.

Ideally, at some point in the future, flow control will be exposed to
the user of the library, but this is currently out of scope.

### DIFF
--- a/src/generic/codec.rs
+++ b/src/generic/codec.rs
@@ -126,7 +126,7 @@ pub struct DecodeBuf<'a> {
 
 #[derive(Debug)]
 pub struct BytesList {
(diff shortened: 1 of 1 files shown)


hawkw:
:hands_raised:

PR #30, opened 2018-01-10
Fix large_unary interop test.

carllerche:
Depends on #29 to fix the hanging issue.

### DIFF
--- a/tower-grpc-interop/travis-interop.sh
+++ b/tower-grpc-interop/travis-interop.sh
@@ -24,4 +24,6 @@ trap 'echo ":; killing test server"; kill ${SERVER_PID};' EXIT
 
 # run the interop test client against the server.
 cargo run -p tower-grpc-interop --bin client -- \
-    --test_case=empty_unary     # this is the only working test case...
+    # Add comma separated test cases
+    --test_case=empty_unary, \
+                large_unary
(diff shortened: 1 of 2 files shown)


tower-grpc-interop/travis-interop.sh
```diff
@@ -24,4 +24,6 @@ trap 'echo ":; killing test server"; kill ${SERVER_PID};' EXIT
 
 # run the interop test client against the server.
 cargo run -p tower-grpc-interop --bin client -- \
-    --test_case=empty_unary     # this is the only working test case...
\ No newline at end of file
+    # Add comma separated test cases
```
hawkw: i _think_ bash will see this comment as "whitespace" after the continuation `\`, i had problems with that when i was writing the script

hawkw: (just moving the comment to before the line starting with `cargo run ...` is probably the easiest solution)

tower-grpc-interop/travis-interop.sh
```diff
@@ -24,4 +24,4 @@ trap 'echo ":; killing test server"; kill ${SERVER_PID};' EXIT
 
 # run the interop test client against the server.
 cargo run -p tower-grpc-interop --bin client -- \
-    --test_case=empty_unary     # this is the only working test case...
\ No newline at end of file
+    --test_case=empty_unary,large_unary
```
hawkw: This LGTM, although I _think_ that if you want to have them on multiple lines, you should be able to say 
```bash
cargo run -p tower-grpc-interop --bin client -- \
  --test_case=empty_unary,\
              large_unary
```

PR #32, opened 2018-01-11
Zero codegen::Scope in ServiceGenerator::finalize

### LINKED ISSUE #31: Use fresh scope for each protobuf file
PR  #10 rewrote service generation to use a single `codegen::Scope`, in order to generate multiple services. However, the same scope appears to be reused when generating multiple _protobuf files_. Obviously, this leads to code duplication & compiler errors.

hawkw:
When `tower-grpc` is invoked on multiple `.proto` files, code generated for one file can make it into multiple output files. This is because the `ServiceGenerator` reuses the file root `codegen::Scope` even after it is `finalize()`ed.

I've changed `ServiceGenerator::finalize()` to replace the scope with a fresh one after outputting the generated code.

Fixes #31 

Signed-off-by: Eliza Weisman <eliza@buoyant.io>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,9 @@ members = [
   "tower-grpc-build",
   "tower-grpc-examples",
   "tower-grpc-interop",
+
+  # For tests
+  "tests/multifile",
 ]
 
 [dependencies]
--- a/tests/multifile/Cargo.toml
+++ b/tests/multifile/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "multifile"
+version = "0.1.0"
+authors = ["Carl Lerche <me@carllerche.com>"]
+publish = false
+
+[dependencies]
+bytes = "0.4"
+prost = { git = "https://github.com/danburkert/prost" }
+prost-derive = { git = "https://github.com/danburkert/prost" }
+tower-grpc = { path = "../../" }
+
+[build-dependencies]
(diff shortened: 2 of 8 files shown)


carllerche:
Thanks for this!

What would be a good strategy to add a test for this change so we avoid future breakage?

hawkw:
@carllerche hmm, I think a good starting point might just be including an example that uses more than one `.proto` file. I'm pretty sure that's why we never caught this issue earlier.

carllerche:
@hawkw sounds like a good option. Are you able to include that? We can then ensure all the examples get compiled as part of CI.

hawkw:
@carllerche sure, I can do that!

hawkw:
This issue only occurs when there are multiple _output_ files; the interop test protos don't catch this, because there are multiple input proto files that are compiled to one output Rust file.

hawkw:
There is [a comment](https://github.com/danburkert/prost/blob/2bcb494654ef1885c13eb234e0e235eae34b7999/prost-build/src/lib.rs#L186-L187) in the `prost-build` docs that indicates why this is necessary.

carllerche:
What's the status of this? Is there anything that I need to do?

hawkw:
Thanks for adding the test @carllerche! Just to confirm, that currently fails to compile on `master`?

carllerche:
Yes, I checked it failed on master.

PR #37, opened 2018-01-18
Handle proto import collisions (eliza edition)

hawkw:


carllerche:
Should we close this?

hawkw:
@carllerche yeah, I think this is no longer necessary.

PR #38, opened 2018-01-19
Fix more complicated namespace codegen

carllerche:
Includes a test that uses fancier namespacing.


carllerche:
This also gives up on trying to add `use` statements for protobuf message types from other modules. There are too many edge cases, and it didnt' seem worth dealing with it.

PR #39, opened 2018-01-22
Fix client server streaming codgen

carllerche:
The existing codegen included an unused generic, which prevents calling
the actual function given that the generic can not be inferred.


carllerche:
I'm leaning towards not having a ton of different sub crates for tests... if it gets out of hand, we can split out later.

PR #46, opened 2018-01-29
Zero codegen::Scope in ServiceGenerator::finalize, again 

### LINKED ISSUE #45: prost 0.3 update breaks Conduit generated code
After updating [my Conduit branch that depends on `tower-grpc` from this repo](https://github.com/runconduit/conduit/pull/225) to depend on `prost` 0.3.0, I see the following compile error:

```
   Compiling conduit-proxy v0.1.3 (file:///Users/eliza/Code/go/src/github.com/runconduit/conduit/proxy)
error[E0432]: unresolved import `super::Update`
   --> /Users/eliza/Code/go/src/github.com/runconduit/conduit/target/debug/build/conduit-proxy-e25c6e4edf2849f0/out/conduit.proxy.telemetry.rs:147:17
    |
147 |     use super::{Update, ReportRequest, ReportResponse};
    |                 ^^^^^^ no `Update` in `control::pb::proxy::telemetry`

error: aborting due to previous error

error: Could not compile `conduit-proxy`.

```

hawkw:
It looks like the code added in PR #32 to fix the issue where `ServiceGenerator`s would output duplicate code when generating code for multiple .proto files (issue #31) was [accidentally removed](https://github.com/tower-rs/tower-grpc/commit/bbf1faca7e6e04aa611117e83d318dae44fc3835#diff-d15ac6177e75577c93c4041346f98ed0L125) in #44, creating a regression that caused issue #45.  

I've added this code back.

Furthermore, I was curious as to why the `test/multifile` test crate, which was added specifically to test for this bug, didn't catch this regression. It turns out that `test/multifile` does in fact fail to build on `master`, but we don't actually try to build it on Travis. Therefore, I also added a step to the CI build that tries to build the `test/multifile` and `test/collide` code-generation test crates, so we don't merge code that breaks these cases in the future (39e73acacd1f802c0e5968295a86e69d19e82a2a).

Fixes #45.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ matrix:
 
 script:
   - cargo test
+  # build test crates
+  - cargo build -p multifile
+  - cargo build -p collide
   # run interop tests
   - ./tower-grpc-interop/travis-interop.sh
 
--- a/tower-grpc-build/src/client.rs
+++ b/tower-grpc-build/src/client.rs
@@ -1,6 +1,5 @@
 use codegen;
 use prost_build;
-use std::fmt;
 
 /// Generates service code
 pub struct ServiceGenerator;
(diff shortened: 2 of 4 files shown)


hawkw:
Should also note that I tested building Conduit with a local override in `.cargo/config`, and confirmed that these changes do fix the regression :)

PR #48, opened 2018-02-06
Fix various issues related to name case transformations

### LINKED ISSUE #47: codegen: case transformations are not applied uniformly
It looks like there are case transformations applied by prost/t-grpc-build on input protobuf definition, however name mangling doesn't seem to be applied uniformly, resulting in build failures.

A sample case would be the [rkt metadata API](https://github.com/rkt/rkt/blob/v1.29.0/api/v1alpha/api.proto#L460), which contains a service named `PublicAPI` (NB: this follows Go naming convention).

When processing that through prost/t-grpc-build, the generated code has a mix of `API` and `Api`, such as:
```
impl<T> PublicApiServer<T>
where T: PublicApi,
 { ... }

impl<T> tower::Service for PublicApiServer<T>
where T: PublicAPI,
 { ... }

fn call(&mut self, request: Self::Request) -> Self::Future {
  match request.uri().path() {                                
    "/v1alpha.PublicAPI/GetInfo" => { ... }
  ...
}      
```

hawkw:
Fixes #47 

There are currently a number of issues in how `tower-grpc-build` handles name case transformations on identifiers defined in protobuf. 

I've rewritten the code generator for services to always use the `prost-build` `service.name` field rather than`service.proto_name`, which fixes #47. I've also fixed similar issues that occur when method names include acronyms (or acronym-like strings...), and changed the server code gen to use _exclusively_ Rust-style identifiers, rather than using `method.proto_name` for internal types.

I've added a new test crate, `tests/name-case` to reproduce the issue described in #47 and other similar issues. This crate fails to compile against the current master branch.


.travis.yml
```diff
@@ -17,6 +17,7 @@ script:
   # build test crates
   - cargo build -p multifile
```
carllerche: I think all these lines could be replaced with `cargo build --all`.

tower-grpc-build/Cargo.toml
```diff
@@ -6,3 +6,4 @@ authors = ["Carl Lerche <me@carllerche.com>"]
 [dependencies]
 codegen = { git = "https://github.com/carllerche/codegen" }
 prost-build = "0.3"
+heck = "0.3"
```
carllerche: newline missing?

.travis.yml
hawkw: Didn't know we could do that!

tower-grpc-build/src/lib.rs
```diff
@@ -176,3 +179,25 @@ fn unqualified(ty: &str, level: usize) -> String {
...
+/// identifier.
+///
+/// This is identical to the same [function] in `prost-build`, however, we
+/// reproduce it here as `prost` does not publically export it.
+///
+/// We need this as `prost-build` will only give us the snake-case transformed
+/// names for gRPC methods, but we need to use method names in types as well.
+///
+/// [function]: https://github.com/danburkert/prost/blob/d3b971ccd90df35d16069753d52289c0c85014e4/prost-build/src/ident.rs#L28-L38
+fn to_upper_camel(s: &str) -> String {
```
carllerche: Having to implement this ourselves is unfortunate as it makes us brittle to changes in prost.

Maybe prost could include additional variants in the structs it provides?

Thoughts @danburkert?

hawkw:
@carllerche 
> It would be nice if we didn't have to implement to_upper_camel, but I don't think we have a choice right now.

Totally agree. If `prost` were to expose this API later, that definitely can be removed.

tower-grpc-build/src/lib.rs
danburkert: I don't think this is particularly brittle WRT changes in prost, since prost never exports a `to_upper_camel`ized version of a method name.  This appears to be completely specific to grpc, so I'm not in favor of exporting it from prost-build.  Perhaps I'm missing something, though.

PR #53, opened 2018-03-16
Distinguish protocol errors from server errors

rbtying:
It's often helpful for debugging and error handling to be able to distinguish between `Status::Unknown` from a server response and `Status::Unknown` due to something in the middle. Also, other header/trailer values are often useful for application logic.

### DIFF
--- a/src/client/client_streaming.rs
+++ b/src/client/client_streaming.rs
@@ -5,6 +5,7 @@ use futures::{Future, Stream, Poll};
 use http::{response, Response};
 use prost::Message;
 use tower_h2::{Body, Data};
+use error::ProtocolError;
 
 #[derive(Debug)]
 pub struct ResponseFuture<T, U, B> {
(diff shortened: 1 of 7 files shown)


src/client/client_streaming.rs
```diff
@@ -46,17 +47,16 @@ where T: Message + Default,
...
+                            ::Error::Inner(()) => ::Error::Protocol(ProtocolError::Internal),
+                            ::Error::Decode(e) => ::Error::Decode(e),
+                            ::Error::Grpc(s, h) => ::Error::Grpc(s, h),
                         });
 
                     let message = match try_ready!(res) {
                         Some(message) => message,
-                        // TODO: handle missing message
-                        None => unimplemented!(),
+                        None => return Err(::Error::Protocol(ProtocolError::MissingMessage)),
```
seanmonstar: Oh yikes, nice!

src/generic/codec.rs
```diff
@@ -239,11 +241,11 @@ where T: Decoder,
                 0 => false,
                 1 => {
                     trace!("message compressed, compression not supported yet");
-                    return Err(Status::UNIMPLEMENTED);
+                    return Err(::Error::Protocol(ProtocolError::Unsupported("compression not supported yet")));
```
seanmonstar: Any reason these shouldn't be specific variants also?

src/generic/codec.rs
```diff
@@ -239,11 +241,11 @@ where T: Decoder,
...
                 },
                 _ => {
                     trace!("unexpected compression flag");
-                    return Err(Status::UNKNOWN);
+                    return Err(::Error::Protocol(ProtocolError::Unsupported("unexpected compression flag")));
```
seanmonstar: This is usually just plain wrong, right? Not something we don't support yet...

src/generic/codec.rs
rbtying: No reason in particular, I can pull them up a level!

src/generic/codec.rs
rbtying: I'm actually not too sure. I'll pull it up a level, but it might not be something that the caller would want to distinguish?

rbtying:
Thanks for the review!

src/generic/codec.rs
seanmonstar: Sorry, I meant variants on `ProtocolError`, not on the top level `Error`.

carllerche:
I like this!

@seanmonstar, looks like you have the review covered. Feel free to merge when you are happy.

PR #60, opened 2018-04-24
Fix unused imports in generated code

### LINKED ISSUE #56: unused imports in server code
```
   Compiling conduit-proxy-controller-grpc v0.3.0 (file:///Users/ver/go/src/github.com/runconduit/conduit/proxy/controller-grpc)
warning: unused import: `Update`
   --> /Users/ver/go/src/github.com/runconduit/conduit/target/debug/build/conduit-proxy-controller-grpc-3203dc03b138711d/out/conduit.proxy.destination.rs:247:45
    |
247 |             use super::super::{Destination, Update};
    |                                             ^^^^^^
    |
    = note: #[warn(unused_imports)] on by default
```

offending code:
```rust
        pub mod methods {
            use ::tower_grpc::codegen::server::*;
            use super::super::{Destination, Update};

            pub struct Get<T>(pub T);

            impl<T> tower::ReadyService for Get<T>
            where T: Destination,
            {
                type Request = grpc::Request<super::super::super::super::super::common::Destination>;
                type Response = grpc::Response<T::GetStream>;
                type Error = grpc::Error;
                type Future = T::GetFuture;

                fn call(&mut self, request: Self::Request) -> Self::Future {
                    self.0.get(request)
                }
            }
        }
    }
```

hawkw:
Fixes #56 

Previously, `tower-grpc-build`'s code generation for streaming RPCs may generate unused import statements, which cause the compiler to output a warning. This can occur in the generated server code's `methods` module for server-streaming RPCs, and in the generated client module for client-streaming RPCs, with the client code importing the method's request type and the server code importing the method's response type when. these types are not needed.

I've modified the generation of import statements to check if the RPC is streaming and skip the import of the appropriate type if so. Previously, imports were generated in a `for` loop that looped over the method's input and output types in a slice, in order to reduce repetition. This was no longer possible, as this change required different behaviour for the input and output types. Instead, I've factored out the repeated code into a trait `ImportType` (implemented for `codegen::Scope` and `codegen::Module`) which is used for both the client and server code generation.

In addition, I've added a new test crate, called `unused-imports`, to reproduce this bug (in the server streaming, client streaming, and bidirectional streaming cases). The test crate is compiled with `#![deny(warnings)]` so that regressions will appear as test failures.

### DIFF
--- a/tower-grpc-build/src/lib.rs
+++ b/tower-grpc-build/src/lib.rs
@@ -119,6 +119,29 @@ impl prost_build::ServiceGenerator for ServiceGenerator {
     }
 }
 
+/// Extension trait for importing generated types into `codegen::Scope`s.
+trait ImportType {
+    fn import_type(&mut self, ty: &str, level: usize);
+}
+
+// ===== impl ImportType =====
+
+impl ImportType for codegen::Scope {
+    fn import_type(&mut self, ty: &str, level: usize) {
+        if !is_imported_type(ty) {
+            let (path, ty) = super_import(ty, level);
+
+            self.import(&path, &ty);
+        }
+    }
+}
+
+impl ImportType for codegen::Module {
+    fn import_type(&mut self, ty: &str, level: usize) {
+        self.scope().import_type(ty, level);
+    }
+}
+
 // ===== utility fns =====
 
 fn method_path(service: &prost_build::Service, method: &prost_build::Method) -> String {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = [
   "tests/multifile",
   "tests/collide",
   "tests/name-case",
+  "tests/unused-imports",
 ]
 
 [dependencies]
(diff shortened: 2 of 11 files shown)


tower-grpc-build/src/lib.rs
```diff
@@ -119,6 +119,29 @@ impl prost_build::ServiceGenerator for ServiceGenerator {
     }
 }
 
+/// Extension trait for importing generated types into `codegen::Scope`s.
+trait ImportType {
```
carllerche: I'm not sure I follow why a trait is needed instead of inherent fns on `Scope` and `Module`.

hawkw: @carllerche we can't define inherent impls for `Scope` and `Module` here, since they come from the `codegen` crate; and I didn't want to add any of this to `codegen`, since it's _very_ `tower-grpc`-specific. 

It probably could have just been bare functions like  `fn import_type(module: &mut Module, ty: &str, level: usize)`, but then they'd have to have different names since they can't be overloaded for `Module` and `Scope`, and I thought having the `Module`/`Scope` in the method-receiver position communicated the intent more clearly --- it's the module/scope that imports the type...

carllerche: Ah yes, I forgot it was in a different crate.

PR #62, opened 2018-05-07
Fix deprecated usage of put_u32/get_u32

repi:
Usage of old `put_u32::<BigEndian>` / `get_u32::<BigEndian>` is deprecated in favor of `put_u32_be` / `get_u32_be`.

This generated compile error on latest on nightly


seanmonstar:
Thanks!

PR #70, opened 2018-07-12
Set Content-Type for server responses

plumenator:
The associated constant `generic::codec::Codec::CONTENT_TYPE` is currently
unused and it's not clear how it can be used to set HTTP headers.

This change moves the constant to `generic::codec::Encoder::CONTENT_TYPE` and sets the header in server responses.

### DIFF
--- a/src/generic/codec.rs
+++ b/src/generic/codec.rs
@@ -13,11 +13,6 @@ use error::ProtocolError;
 
 /// Encodes and decodes gRPC message types
 pub trait Codec {
-    /// The content-type header for messages using this encoding.
-    ///
-    /// Should be `application/grpc+yourencoding`.
-    const CONTENT_TYPE: &'static str;
-
     /// The encode type
     type Encode;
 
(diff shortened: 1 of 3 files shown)


src/generic/codec.rs
```diff
@@ -12,11 +12,6 @@ use error::ProtocolError;
 
 /// Encodes and decodes gRPC message types
 pub trait Codec {
-    /// The content-type header for messages using this encoding.
-    ///
-    /// Should be `application/grpc+yourencoding`.
-    const CONTENT_TYPE: &'static str;
```
hawkw: Perhaps if this was replaced with a separate associated const on the `Encoder` trait, we could continue using it rather than hard-coding it?

src/generic/server/streaming.rs
```diff
@@ -56,7 +57,15 @@ where T: Future<Item = Response<S>,
         let response = response.into_http();
 
         // Map the response body
-        let (head, body) = response.into_parts();
+        let (mut head, body) = response.into_parts();
+
+        // Set the content type
+        // TODO: Don't hard code this here
+        let content_type = "application/grpc+proto";
```
hawkw: If the `CONTENT_TYPE` associated const moved to the `Encoder` trait, I think we could say
```rust
head.headers.insert(
    header::CONTENT_TYPE,
    header::HeaderValue::from_static(content_type),
);
```
instead...

src/generic/codec.rs
plumenator: Done

carllerche:
@hawkw LGTM, so feel free to merge if you like it.

PR #78, opened 2018-08-01
Basic support for Status error messages

### LINKED ISSUE #77: Status doesn't take a message String
It looks like `tower_grpc::Status` currently only encapsulates the status code enum, and not the optional String description. Is this an intentional decision, or a haven't-gotten-around-to-it? Either way, is this a big scary thing, or simple a matter of adding a field and serializing the bytes in the right place?

The existence of https://github.com/tower-rs/tower-grpc/issues/28 suggests that this is intended to be addressed at some point. Any pointers/warnings before I jump in and add it? :)

Thanks!

illicitonion:
Fixes #77

### DIFF
--- a/src/status.rs
+++ b/src/status.rs
@@ -1,205 +1,112 @@
-use std::fmt;
-
+use bytes::Bytes;
 use h2;
-use http::header::{HeaderMap, HeaderValue};
-use http::status::StatusCode;
+use http::{self, HeaderMap};
+use http::header::HeaderValue;
+use std::fmt::Display;
+use std::io;
 
 #[derive(Debug, Clone)]
 pub struct Status {
     code: Code,
+    error_message: String,
+    binary_error_details: Bytes,
 }
 
-#[derive(Clone, Copy, PartialEq, Eq)]
-pub struct Code(Code_);
+const GRPC_STATUS_HEADER_CODE: &str = "grpc-status";
+const GRPC_STATUS_MESSAGE_HEADER: &str = "grpc-message";
+const GRPC_STATUS_DETAILS_HEADER: &str = "grpc-status-details-bin";
 
 impl Status {
-    #[inline]
+    pub fn with_code(code: Code) -> Status {
+        Status {
+            code,
+            error_message: String::new(),
+            binary_error_details: Bytes::new(),
+        }
+    }
+
+    pub fn from_header_map(header_map: &HeaderMap) -> Option<Status> {
+        header_map.get(GRPC_STATUS_HEADER_CODE).clone().map(|code| {
+            let code = ::Code::from_bytes(code.as_ref());
+            let error_message = header_map.get(GRPC_STATUS_HEADER_CODE)
+                .map(|header| header.to_str().map(|header| header.to_owned()))
+                .unwrap_or_else(|| Ok(String::new()));
+            let binary_error_details = header_map.get(GRPC_STATUS_DETAILS_HEADER)
+                .map(|h| Bytes::from(h.as_bytes())).unwrap_or_else(Bytes::new);
+            match error_message {
+                Ok(error_message) => Status {
+                    code,
+                    error_message,
+                    binary_error_details,
+                },
+                Err(err) => {
+                    warn!("Error deserializing status message header: {}", err);
+                    Status {
+                        code: Code::Unknown,
+                        error_message: format!("Error deserializing status message header: {}", err),
+                        binary_error_details,
+                    }
+                }
+            }
+        })
+    }
+
     pub fn code(&self) -> Code {
         self.code
     }
 
-    pub const OK: Status = Status {
-        code: Code(Code_::Ok),
-    };
+    pub fn error_message(&self) -> &str {
+        &self.error_message
+    }
 
-    #[deprecated(note = "use Status::CANCELLED")]
-    pub const CANCELED: Status = Status::CANCELLED;
-
-    pub const CANCELLED: Status = Status {
-        code: Code(Code_::Cancelled),
-    };
-
-    pub const UNKNOWN: Status = Status {
-        code: Code(Code_::Unknown),
-    };
-
-    pub const INVALID_ARGUMENT: Status = Status {
-        code: Code(Code_::InvalidArgument),
-    };
-
-    pub const DEADLINE_EXCEEDED: Status = Status {
-        code: Code(Code_::DeadlineExceeded),
-    };
-
-    pub const NOT_FOUND: Status = Status {
-        code: Code(Code_::NotFound),
-    };
-
-    pub const ALREADY_EXISTS: Status = Status {
-        code: Code(Code_::AlreadyExists),
-    };
-
-    pub const PERMISSION_DENIED: Status = Status {
-        code: Code(Code_::PermissionDenied),
-    };
-
-    pub const RESOURCE_EXHAUSTED: Status = Status {
-        code: Code(Code_::ResourceExhausted),
-    };
-
-    pub const FAILED_PRECONDITION: Status = Status {
-        code: Code(Code_::FailedPrecondition),
-    };
-
-    pub const ABORTED: Status = Status {
-        code: Code(Code_::Aborted),
-    };
-
-    pub const OUT_OF_RANGE: Status = Status {
-        code: Code(Code_::OutOfRange),
-    };
-
-    pub const UNIMPLEMENTED: Status = Status {
-        code: Code(Code_::Unimplemented),
-    };
-
-    pub const INTERNAL: Status = Status {
-        code: Code(Code_::Internal),
-    };
-
-    pub const UNAVAILABLE: Status = Status {
-        code: Code(Code_::Unavailable),
-    };
-
-    pub const DATA_LOSS: Status = Status {
-        code: Code(Code_::DataLoss),
-    };
-
-    pub const UNAUTHENTICATED: Status = Status {
-        code: Code(Code_::Unauthenticated),
-    };
-
-    pub(crate) fn from_bytes(bytes: &[u8]) -> Status {
-        let code = match bytes.len() {
-            1 => {
-                match bytes[0] {
-                    b'0' => Code_::Ok,
-                    b'1' => Code_::Cancelled,
-                    b'2' => Code_::Unknown,
-                    b'3' => Code_::InvalidArgument,
-                    b'4' => Code_::DeadlineExceeded,
-                    b'5' => Code_::NotFound,
-                    b'6' => Code_::AlreadyExists,
-                    b'7' => Code_::PermissionDenied,
-                    b'8' => Code_::ResourceExhausted,
-                    b'9' => Code_::FailedPrecondition,
-                    _ => return Status::parse_err(),
-                }
-            },
-            2 => {
-                match (bytes[0], bytes[1]) {
-                    (b'1', b'0') => Code_::Aborted,
-                    (b'1', b'1') => Code_::OutOfRange,
-                    (b'1', b'2') => Code_::Unimplemented,
-                    (b'1', b'3') => Code_::Internal,
-                    (b'1', b'4') => Code_::Unavailable,
-                    (b'1', b'5') => Code_::DataLoss,
-                    (b'1', b'6') => Code_::Unauthenticated,
-                    _ => return Status::parse_err(),
-                }
-            },
-            _ => return Status::parse_err(),
-        };
-
-        Status::new(Code(code))
+    pub fn binary_error_details(&self) -> &Bytes {
+        &self.binary_error_details
     }
 
     // TODO: It would be nice for this not to be public
-    pub fn to_header_value(&self) -> HeaderValue {
-        match self.code.0 {
-            Code_::Ok => HeaderValue::from_static("0"),
-            Code_::Cancelled => HeaderValue::from_static("1"),
-            Code_::Unknown => HeaderValue::from_static("2"),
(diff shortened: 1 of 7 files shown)


illicitonion:
I believe this is sufficient for the client to work. Very open to discussions about what the API should look like, and what testing would be useful. I'm also interested in people's thoughts on how non-ASCII values should be handled; right now I just use `String::from_utf8_lossy` to preserve as much information as possible, which seems reasonable...

I tested this by starting a trivial server served with pingcap's grpcio which returns an RpcStatus with a message, and running a trivial client.

For making the server side work, I propose that `to_header_value` return a tuple of `(HeaderValue, Option<HeaderValue>)`. It would be nice for this to not have to return a `Result`, but technically constructing a `HeaderValue` from a `str` or `[u8]` may fail... I guess it could just return a `Result`... Any thoughts on that API would also be very welcome :)

per-gron:
How does this added field relate to the `Error` enum in `src/error.rs`? That data type contains error messages as well as the status (the headers in the `Grpc` variant has this data).

I'm not saying I think this is a bad change; I think the current `Error` type is a hassle to use, but to me it seems like the premise of #77 (you can't get the error message) is incorrect, it's just not very convenient at the moment.

I think it would be good to have some kind of plan for how to deal not only with the error message but also the binary error details that the [C gRPC library has](https://github.com/grpc/grpc/blob/master/include/grpcpp/impl/codegen/status.h), as well as how to simplify the `Error` type.

As this PR stands today it seems like it duplicates data which is not ideal. I suggest to take this idea even further (not necessarily in this PR):

* Also add a binary error detail field
* Remove the current `Error` type and just use `Status` instead: The `Grpc` variant of that enum should not need to have headers once the error headers are in `Status` (this is how the C gRPC API is), the `Protocol` variant of that enum is in my opinion not needed; there is official documentation for how to map this kind of errors to a `Status` and that is how the other gRPC libraries work. The same goes for the `Decode` variant but that has yet another reason to be removed: It currently (and unnecessarily) ties tower-grpc to prost. Removing the `Inner` variant would also be a boon because at the moment the client and server `Error` types are not the same which is just a hassle to deal with in practical code for very little benefit that I can see.

illicitonion:
I'm definitely happy to expand this to clean up the `Error`/`Status` divide :) I've only spend a few hours prototyping switching a crate I work on from grpcio to tower-grpc, and don't have much knowledge of either grpc or tower-grpc, so than you for raising these questions with more context!

So, two big (orthogonal) questions:
1. Should `Error::Grpc` have a `HeaderMap` as a public field? My instinct here is no. Given that, should it a) have it as a private field, and expose the `String`s (or maybe `Bytes`) as references to the `HeaderMap`, b) take the `Bytes` out of the `HeaderMap`, c) clone them out? @per-gron's comment suggests b.
2. Is there a need for `Error` as separate from `Status`? As @per-gron suggests, `Protocol` can probably be obsoleted with https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md, and `Decode` similarly... What about `Inner`?

src/status.rs
```diff
@@ -1,166 +1,156 @@
...
-        code: Code(Code_::ResourceExhausted),
-    };
-
-    pub const FAILED_PRECONDITION: Status = Status {
-        code: Code(Code_::FailedPrecondition),
-    };
+    pub fn from_header_map(header_map: &mut HeaderMap) -> Option<Status> {
+        if let Some(code) = header_map.remove("grpc-status") {
+            let code = ::Code::from_bytes(code.as_ref());
+            let error_message = header_map.remove("grpc-message").map(|h| h.into());
```
hawkw: Again, take it or leave it: `.map(|h| h.into());` could also be written as just `.map(HeaderValue::into)` (or even `.map(Into::into)`, which might be a little more terse.

src/status.rs
```diff
@@ -1,166 +1,156 @@
...
-
-    pub const RESOURCE_EXHAUSTED: Status = Status {
-        code: Code(Code_::ResourceExhausted),
-    };
-
-    pub const FAILED_PRECONDITION: Status = Status {
-        code: Code(Code_::FailedPrecondition),
-    };
+    pub fn from_header_map(header_map: &mut HeaderMap) -> Option<Status> {
+        if let Some(code) = header_map.remove("grpc-status") {
```
hawkw: Take it or leave it: 
```rust
if let Some(code) = header_map.remove("grpc-status") {
    // ...
    Some(Status {
        // ...
    })
} else {
    None
}
```
could be simplified as
```rust
header_map.remove("grpc-status").map(|code| {
    // ...
    Status {
        // ...
    }
})
```

src/status.rs
```diff
@@ -1,166 +1,156 @@
...
-    };
-
-    pub const RESOURCE_EXHAUSTED: Status = Status {
-        code: Code(Code_::ResourceExhausted),
-    };
-
-    pub const FAILED_PRECONDITION: Status = Status {
-        code: Code(Code_::FailedPrecondition),
-    };
+    pub fn from_header_map(header_map: &mut HeaderMap) -> Option<Status> {
```
hawkw: I'm not sure how I feel about this function taking the header map mutably, and removing the headers from it...that seems somewhat surprising. If I were to use that `HeaderMap` again later, I'd expect all of its headers to still be intact. Perhaps this should take the header map by an immutable reference, and clone the appropriate header values?

Admittedly, removing them from the map may be more performant, but it seems like it could lead to bugs...

src/status.rs
```diff
@@ -1,166 +1,156 @@
...
+        if let Some(code) = header_map.remove("grpc-status") {
+            let code = ::Code::from_bytes(code.as_ref());
+            let error_message = header_map.remove("grpc-message").map(|h| h.into());
+            let binary_error_details = header_map.remove("grpc-status-details-bin").map(|h| h.into());
```
hawkw: Another nit: the string literals `"grpc-message"`, `"grpc-status"`, and `"grpc-status-details-bin"` show up several times in this PR. It would be nice if these could be constants...

tower-grpc-build/src/server.rs
```diff
@@ -387,9 +387,8 @@ macro_rules! try_ready {
 
         let mut poll_trailers_catch_all = codegen::Block::new("Err(ref status) =>");
         poll_trailers_catch_all
-            .line("let mut map = http::HeaderMap::new();")
-            .line("map.insert(\"grpc-status\", status.to_header_value());")
-            .line("Ok(Some(map).into())")
+            .line("let map = status.to_header_map();")
+            .line("Ok(Some(map?).into())")
```
hawkw: This might be clearer as 
```rust
status.to_header_map().map(|map| Some(map).into())
```
instead of using `?` from inside of an `Ok(...)` expression...

tower-grpc-build/src/server.rs
```diff
@@ -387,9 +387,8 @@ macro_rules! try_ready {
 
         let mut poll_trailers_catch_all = codegen::Block::new("Err(ref status) =>");
         poll_trailers_catch_all
-            .line("let mut map = http::HeaderMap::new();")
-            .line("map.insert(\"grpc-status\", status.to_header_value());")
-            .line("Ok(Some(map).into())")
+            .line("let map = status.to_header_map();")
```
hawkw: :+1: this is a much nicer API!

hawkw:
Also, it looks like there are some conflicts between this branch and `master` --- @illicitonion, can I get you to resolve those?

src/status.rs
```diff
@@ -1,166 +1,156 @@
...
+            Code::DataLoss => HeaderValue::from_static("15"),
+            Code::Unauthenticated => HeaderValue::from_static("16"),
+        };
 
-    pub const UNAVAILABLE: Status = Status {
-        code: Code(Code_::Unavailable),
-    };
+        header_map.insert("grpc-status", status_header);
+        if let Some(ref error_message) = self.error_message {
+            header_map.insert("grpc-message", HeaderValue::from_shared(error_message.clone()).map_err(invalid_header_value_byte_to_h2)?);
```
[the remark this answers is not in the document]
illicitonion: Done

src/status.rs
illicitonion: Done

src/status.rs
illicitonion: Done

src/status.rs
illicitonion: Yeah, I was pretty torn on this... Switched to immutable reference + clone

src/status.rs
illicitonion: Done

tower-grpc-build/src/server.rs
illicitonion: Done

illicitonion:
Merged master and resolved all comments - thanks for the review!

@seanmonstar @carllerche there are some open questions for y'all

src/status.rs
```diff
@@ -1,166 +1,156 @@
...
+fn invalid_header_value_byte_to_h2(err: InvalidHeaderValueBytes) -> h2::Error {
+    debug!("Invalid header: {:?}", err);
+    h2::UserError::MalformedHeaders.into()
+}
 
-    pub const UNAUTHENTICATED: Status = Status {
-        code: Code(Code_::Unauthenticated),
-    };
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Code {
```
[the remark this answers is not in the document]
carllerche: I wouldn't worry about this yet because all enums are public anyway.

Before pushing to crates.io, we would do a pass.

Cargo.toml
```diff
@@ -29,7 +29,8 @@ members = [
 bytes = "0.4.7"
 futures = "0.1"
 http = "0.1"
-h2 = "0.1"
+# Need unstable features for UserError
+h2 = { version = "0.1", features = ["unstable"] }
```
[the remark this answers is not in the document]
illicitonion: Switched to use an io Error, which doesn't quite feel right, but is at least stable...

src/error.rs
```diff
@@ -1,10 +1,9 @@
 use prost::DecodeError;
-use http::HeaderMap;
 use h2;
 
 #[derive(Debug)]
 pub enum Error<T = ()> {
-    Grpc(::Status, HeaderMap),
```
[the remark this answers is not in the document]
illicitonion: Re-added

carllerche:
I merged master, but now some tests fail.

src/status.rs
```diff
@@ -1,167 +1,162 @@
...
-        self.code
+    pub fn with_code(code: Code) -> Status {
+        Status {
+            code,
+            error_message: None,
+            binary_error_details: None,
+        }
     }
 
-    pub const OK: Status = Status {
```
carllerche: What was the reason to remove this? It seems convenient as a way to get quick status codes?

illicitonion: `Code` is now a public enum, so can easily be gotten as needed

illicitonion:
Fixed up the merge, CI is now green :)

carllerche:
@per-gron Would you mind reviewing this?

src/status.rs
```diff
@@ -1,167 +1,162 @@
...
-use http::header::{HeaderMap, HeaderValue};
-use http::status::StatusCode;
+use http::{self, HeaderMap};
+use http::header::{HeaderValue, InvalidHeaderValueBytes};
+use std::io;
 
 #[derive(Debug, Clone)]
 pub struct Status {
     code: Code,
+    error_message: Option<Bytes>,
```
per-gron: I think it would be better if `error_message` was a `String`. In this comment I argue for changing `Bytes` => `String`, in the comment below I argue for removing `Option`.

The error message field is supposed to be human readable. Storing them as `Bytes` makes this considerably more painful, just for handling the very rare corner case of non-UTF-8 error messages. Java and C# gRPC store this message as a (unicode) String. Java does percent encoding of invalid characters, [C# fails nongracefully to parse invalid UTF-8](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/Internal/AsyncCall.cs#L119) (ping @jtattermusch is this true?).

I think the pain of this not being a String vs the value of supporting arbitrary binary error messages is too large to justify `error_message` being `Bytes`.

https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/Internal/BatchContextSafeHandle.cs#L87
https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/Status.java#L353

src/status.rs
```diff
@@ -1,167 +1,162 @@
...
+    binary_error_details: Option<Bytes>,
```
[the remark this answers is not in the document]
illicitonion: Switched to empty `Bytes` and `String` respectively

src/status.rs
illicitonion: Yeah, I went back and forth on this myself; I went with `Bytes` mostly because there isn't an obviously good place to complain that it was invalid if it was, but have made it a `String` because that's probably more useful.

per-gron: Thanks for updating the code!

jtattermusch: yes, the error message should be human-readable string (utf8 by default). If one wants to send additional error details that are binary, the right way is to add a binary-values trailer with extra debug info.  According to best practices, services can attach protobuf-serialized error status as an extra trailer. https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto

illicitonion:
Rebased again... Any chance we can get this merged, if people are happy with it? :) Thanks!

per-gron:
Thanks for your contribution! I'm not on the Buoyant team but since I'm not the only one who has approved this PR I'll go ahead and merge.

Issue #79, opened 2018-08-10
Metadata support


D4nte:
Hi,

If I understand correctly, metadata support (e.g. for authentication using macaroons) is not yet supported?

Is there an issue to track the implementation support so I can keep an eye on it?

Thanks!

D4nte:
So apparently custom-metadata is an [HTTP2 thing](https://godoc.org/google.golang.org/grpc/metadata) I wonder if it's possible to set the metadata in the tower when setting up the HTTP connection.

seanmonstar:
Yea, metadata is sent via headers. The underlying service could be used to add or read metadata, in that they can do anything with headers. It's probably also a good idea to have metadata helper methods in this crate, since there's some rules around binary encoded metadata headers for grpc.

D4nte:
Ok I was able to set the header manually.

See #80.

I am guessing the rule you are mentioning @seanmonstar is that binary header should be encoded in base64?

I could add a binary example to #80 if you think it's worth it.

seanmonstar:
There is now extensive metadata support for requests and responses.

PR #80, opened 2018-08-10
Adding custom metadata in helloworld client example

D4nte:
In relation to #79. While it may be straightforward to some that gRPC metadata can directly be set as custom headers in the HTTP2 layer, it was not to me.

I think that adding it in the example may be useful to some until the `metadata` check box is ticked.

### DIFF
--- a/tower-grpc-examples/src/metadata/server.rs
+++ b/tower-grpc-examples/src/metadata/server.rs
@@ -1,0 +1,78 @@
+extern crate bytes;
+extern crate env_logger;
+extern crate futures;
+#[macro_use]
+extern crate log;
+extern crate prost;
+#[macro_use]
+extern crate prost_derive;
+extern crate http;
+extern crate tokio_core;
+extern crate tower_grpc;
+extern crate tower_h2;
+
+pub mod metadata {
+    include!(concat!(env!("OUT_DIR"), "/metadata.rs"));
+}
+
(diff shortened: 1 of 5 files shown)


tower-grpc-examples/src/metadata/server.rs
```diff
@@ -0,0 +1,74 @@
...
+#[derive(Clone, Debug)]
+struct Door;
+
+impl server::Doorman for Door {
+    type AskToEnterFuture = future::FutureResult<Response<EnterReply>, tower_grpc::Error>;
+
+    fn ask_to_enter(&mut self, request: Request<EnterRequest>) -> Self::AskToEnterFuture {
+        println!("REQUEST = {:?}", request);
+
+        let message = match request.headers().get("metadata").and_then(|header| header.to_str().ok()) {
```
hawkw: Nit: would be nice if this was wrapped, or broken into something like:
```rust
let metadata = request.headers()
    .get("metadata")
    .and_then(|header| header.to_str().ok());
let message = match metadata {
    // ...
};
```

D4nte: haha, I had that at first and then change it to the one liner...

D4nte:
@hawkw all done, however I do not merge rights despite the approval.

hawkw:
@D4nte sorry about that, I just merged it. Thanks for the contribution!

PR #83, opened 2018-09-11
GRPC status should check against HTTP status code if not otherwise provided

### LINKED ISSUE #82: Interpret HTTP status code from underlying H2 layer if grpc-status is not provided.
https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md

If the grpc-status code is not provided, we should infer one from the HTTP status code.

rbtying:
Fixes #82 .

### DIFF
--- a/src/client/streaming.rs
+++ b/src/client/streaming.rs
@@ -5,6 +5,7 @@ use http::Response;
 use prost::Message;
 use tower_h2::{Body, Data};
 
+use Code;
 use std::marker::PhantomData;
 
 #[derive(Debug)]
@@ -40,11 +41,14 @@ where T: Message + Default,
 
         // Get the response
         let response = try_ready!(response);
+        let status_code = response.status();
 
         // Destructure into the head / body
         let (head, body) = response.into_parts();
 
-        if let Some(status) = super::check_grpc_status(&head.headers) {
+        let status = super::check_grpc_status(&head.headers, status_code);
+
+        if status.code() != Code::OK {
             return Err(::Error::Grpc(status, head.headers));
         }
 
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -6,7 +6,7 @@ pub mod streaming;
 use Status;
 
 use futures::{stream, Stream, Poll};
-use http::{uri, HeaderMap, Uri};
+use http::{uri, HeaderMap, Uri, StatusCode};
 use prost::Message;
 use tower_h2::{HttpService, BoxBody};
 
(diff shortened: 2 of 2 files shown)


src/client/streaming.rs
```diff
@@ -40,11 +41,14 @@ where T: Message + Default,
 
         // Get the response
         let response = try_ready!(response);
+        let status_code = response.status();
 
         // Destructure into the head / body
         let (head, body) = response.into_parts();
 
-        if let Some(status) = super::check_grpc_status(&head.headers) {
```
seanmonstar: I'm just trying to be sure I understand, but it seems like the behavior *before* was that the existence of any `grpc-status` trailer would return an error, always? That seems like a woops...

rbtying:
Yup, that's what would happen. As it turns out grpc-status is usually only
set on error... So this happened to work. Ish.

seanmonstar:
Wait, so then that'd mean that this would be turning cases that were OK into UNKNOWN, right?

seanmonstar:
I remember now. As stated in the spec https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses, it's possible for failures to include a Trailers-Only response. This is when the response headers contain the grpc-status immediately, so that 2 headers frames aren't needed if there would otherwise be no body.

So, in that case, it should only check the headers, and map the status code in trailers.

rbtying:
How do we check the trailers-only case here? What confuses me is that this change (as it stands) actually works in the case where there is a real gRPC body (and I would expect it to fail).

seanmonstar:
If the initial headers include a `grpc-status`, then it is mostly meant as an error, as indicated in the spec. If they aren't there, then the status depends on the final trailers (*or*, if no trailers come, *then* map the original HTTP status code). Does that make sense?

rbtying:
@seanmonstar I implemented roughly the logic we're discussing here in the newest commit on the branch. With local testing it seems to work correctly, although that was also true of the prior diff...

src/client/streaming.rs
```diff
@@ -41,14 +44,19 @@ where T: Message + Default,
...
+        let status_code = response.status();
+
         // Destructure into the head / body
         let (head, body) = response.into_parts();
 
-        if let Some(status) = super::check_grpc_status(&head.headers) {
-            return Err(::Error::Grpc(status, head.headers));
+        // Check the headers for `grpc-status`, in which case we should not parse the body.
+        if let Some(status) = infer_grpc_status(&head.headers, None) {
+            if status.code() != Code::OK {
```
seanmonstar: It'd likely be weird for a `Trailers-Only` response to have `grpc-status = 0`. If the header exists, it's meant to mean a failure and no body... should we assume a status of `OK` to be a mistake?

Issue #90, opened 2018-10-25
Make HTTP a hidden implementation detail, don't expose it in the public API


per-gron:
Right now, tower-grpc exposes to its users that it uses HTTP in a bunch of places. An example is the [`tower_grpc::Error::Grpc`](https://github.com/tower-rs/tower-grpc/blob/master/src/error.rs) enum variant, which exposes the `http::HeaderMap` type from hyperium/http. Another example is the [`tower_grpc::Response`](https://github.com/tower-rs/tower-grpc/blob/master/src/response.rs) type, which has `into_http` and `from_http` methods to convert to and from `http::Response` values.

I think the tower-grpc API should be changed so that no HTTP specific types are visible to its users. There are both philosophical, practical and experience based reasons for this:

# Reasons

## Philosophical argument

gRPC is designed as a protocol that is agnostic to its underlying transport. The [gRPC over HTTP2](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) specification document begins with "This document serves as a detailed description for an implementation of gRPC carried over HTTP2 framing."

It can be argued that it's unlikely that users of tower-grpc are going to ever want any other underlying transport, but we don't really know that. In addition, and this is perhaps the stronger reason, gRPC is a useful and easy to reason about *subset* of HTTP2, which on its own is very complex. By exposing raw HTTP2 constructs to users of tower-grpc, users are unnecessarily exposed to complexities of HTTP2, for little to no benefit.

## Practical argument

The direct exposure of HTTP primitives, at least without helpers which don't exist today, makes it difficult or impossible to properly implement some of the [gRPC language interop tests](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md).

`special_status_message` is about sending an error response with Unicode text, which [gRPC over HTTP2](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) requires to be percent encoded. With tower-grpc's current design, this percent encoding has to be done by end user code, which is very error prone and not necessary. All other gRPC libraries handle this automatically. Even without the complication of the percent encoding, requiring the user to know that you should check the "grpc-message" HTTP header to get the error message is a very unintuitive API.

[...]

PR #92, opened 2018-11-08
Introduce tower_grpc::metadata::MetadataMap type

per-gron:
As per the discussion in #90, it wraps http::HeaderMap to abstract away the fact that tower-grpc is HTTP based. At this time this class is a pure wrapper that does not actually implement any of the gRPC metadata specific stuff (such as binary fields), that will come later.

My goal with this PR is to make `MetadataMap` (largely) equivalent to `HeaderMap` while not exposing http types.

### DIFF
--- a/src/metadata_key.rs
+++ b/src/metadata_key.rs
@@ -34,31 +34,6 @@ impl MetadataKey {
         }
     }
 
-    /// Converts a slice of bytes to a `MetadataKey`.
-    ///
-    /// This function expects the input to only contain lowercase characters.
-    /// This is useful when decoding HTTP/2.0 headers. The HTTP/2.0
-    /// specification requires that all headers be represented in lower case.
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// # use tower_grpc::metadata::*;
-    ///
-    /// // Parsing a lower case metadata key
-    /// let hdr = MetadataKey::from_lowercase(b"content-length").unwrap();
-    /// assert_eq!("content-length", hdr);
-    ///
-    /// // Parsing a metadata key that contains uppercase characters
-    /// assert!(MetadataKey::from_lowercase(b"Content-Length").is_err());
-    /// ```
-    pub fn from_lowercase(src: &[u8]) -> Result<MetadataKey, InvalidMetadataKey> {
-        match HeaderName::from_lowercase(src) {
-            Ok(name) => Ok(MetadataKey { inner: name }),
-            Err(_) => Err(InvalidMetadataKey { _priv: () })
-        }
-    }
-
     /// Converts a static string to a `MetadataKey`.
     ///
     /// This function panics when the static string is a invalid metadata key.
@@ -75,7 +50,7 @@ impl MetadataKey {
     /// // Parsing a metadata key
     /// let CUSTOM_KEY: &'static str = "custom-key";
     /// 
-    /// let a = MetadataKey::from_lowercase(b"custom-key").unwrap();
+    /// let a = MetadataKey::from_bytes(b"custom-key").unwrap();
     /// let b = MetadataKey::from_static(CUSTOM_KEY);
     /// assert_eq!(a, b);
     /// ```
--- a/src/metadata_key.rs
+++ b/src/metadata_key.rs
@@ -1,0 +1,249 @@
+use bytes::Bytes;
+use http;
+use http::header::HeaderName;
+
+use std::borrow::Borrow;
+use std::error::Error;
+use std::fmt;
+use std::str::FromStr;
+
+/// Represents a custom metadata field name.
+///
+/// `MetadataKey` is used as the [`MetadataMap`] key.
+///
+/// [`HeaderMap`]: struct.HeaderMap.html
+#[derive(Clone, Eq, PartialEq, Hash)]
+pub struct MetadataKey {
+    pub(crate) inner: http::header::HeaderName,
+}
+
+/// A possible error when converting a `MetadataKey` from another type.
+#[derive(Debug)]
+pub struct InvalidMetadataKey {
+    _priv: (),
+}
+
+impl MetadataKey {
+    /// Converts a slice of bytes to a `MetadataKey`.
+    ///
+    /// This function normalizes the input.
+    pub fn from_bytes(src: &[u8]) -> Result<MetadataKey, InvalidMetadataKey> {
+        match HeaderName::from_bytes(src) {
+            Ok(name) => Ok(MetadataKey { inner: name }),
+            Err(_) => Err(InvalidMetadataKey::new())
+        }
+    }
+
+    /// Converts a slice of bytes to a `MetadataKey`.
+    ///
+    /// This function expects the input to only contain lowercase characters.
+    /// This is useful when decoding HTTP/2.0 headers. The HTTP/2.0
+    /// specification requires that all headers be represented in lower case.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use tower_grpc::metadata::*;
+    ///
+    /// // Parsing a lower case metadata key
+    /// let hdr = MetadataKey::from_lowercase(b"content-length").unwrap();
+    /// assert_eq!("content-length", hdr);
+    ///
+    /// // Parsing a metadata key that contains uppercase characters
+    /// assert!(MetadataKey::from_lowercase(b"Content-Length").is_err());
+    /// ```
+    pub fn from_lowercase(src: &[u8]) -> Result<MetadataKey, InvalidMetadataKey> {
+        match HeaderName::from_lowercase(src) {
+            Ok(name) => Ok(MetadataKey { inner: name }),
+            Err(_) => Err(InvalidMetadataKey { _priv: () })
+        }
+    }
+
+    /// Converts a static string to a `MetadataKey`.
+    ///
+    /// This function panics when the static string is a invalid metadata key.
+    /// 
+    /// This function requires the static string to only contain lowercase 
+    /// characters, numerals and symbols, as per the HTTP/2.0 specification 
+    /// and header names internal representation within this library.
+    /// 
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use tower_grpc::metadata::*;
+    /// // Parsing a metadata key
+    /// let CUSTOM_KEY: &'static str = "custom-key";
+    /// 
+    /// let a = MetadataKey::from_lowercase(b"custom-key").unwrap();
+    /// let b = MetadataKey::from_static(CUSTOM_KEY);
+    /// assert_eq!(a, b);
+    /// ```
+    /// 
+    /// ```should_panic
+    /// # use tower_grpc::metadata::*;
+    /// #
+    /// // Parsing a metadata key that contains invalid symbols(s):
+    /// MetadataKey::from_static("content{}{}length"); // This line panics!
+    /// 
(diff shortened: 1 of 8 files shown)


seanmonstar:
Wow! I haven't reviewed yet, I'm about to look through, but first thoughts: you really took this seriously, this is impressive!

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    ///
+    /// assert!(map.is_empty());
+    /// assert_eq!(0, map.capacity());
+    /// ```
+    pub fn new() -> Self {
+        MetadataMap::with_capacity(0)
+    }
+
+    /// Convert an HTTP HeaderMap to a MetadataMap
+    pub fn from_headers(mut headers: http::HeaderMap) -> Self {
```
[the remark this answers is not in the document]
per-gron: These two conversion methods are used in `src/request.rs`. Perhaps I should make them `pub(crate)` instead.

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+        let mut map = Self::with_capacity(headers.len());
+        for (name, values) in headers.drain() {
```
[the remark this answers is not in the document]
per-gron: `http::HeaderMap` is equal to `http::HeaderMap<http::HeaderValue>`. `MetadataMap` however has a `http::HeaderMap<MetadataValue>`, so when constructing a `MetadataMap` the values need to be wrapped individually.

`MetadataMap` can't just wrap a `http::HeaderMap<http::HeaderValue>` because there are some important methods that return references to those values, which would be unimplementable then (or would require a separate `MetadataValueRef` type that wrapped `HeaderName&`).

(There is a similar issue for methods that return references to map keys but I worked around that by returning `&str`s instead, which is basically what `HeaderName` is anyways after construction and you don't need to quickly compare them with standard HTTP header names.)

At least this code just moves all the headers, so runtime is proportional to header count but not size.

src/metadata_key.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// # use tower_grpc::metadata::*;
+    ///
+    /// // Parsing a lower case metadata key
+    /// let hdr = MetadataKey::from_lowercase(b"content-length").unwrap();
+    /// assert_eq!("content-length", hdr);
+    ///
+    /// // Parsing a metadata key that contains uppercase characters
+    /// assert!(MetadataKey::from_lowercase(b"Content-Length").is_err());
+    /// ```
+    pub fn from_lowercase(src: &[u8]) -> Result<MetadataKey, InvalidMetadataKey> {
```
carllerche: I would delete this for now. It was added to better support HTTP/2.0 implementations. gRPC should not need it.

src/metadata_map.rs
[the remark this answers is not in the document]
carllerche: I mean... it doesn't seem terrible to have conversions. Being able to easily convert between the two does not violate "hiding".

src/metadata_map.rs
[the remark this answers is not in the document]
carllerche: We could still do it, though it would require an unsafe transmute. stdlib relies on this being valid in a number of places. For example, `Path::new` where `Path(OsStr)`:

```rust
    pub fn new<S: AsRef<OsStr> + ?Sized>(s: &S) -> &Path {
        unsafe { &*(s.as_ref() as *const OsStr as *const Path) }
    }
```

I would use this strategy personally. I think a zero-cost `MetadataMap` wrapper is critical here, otherwise switching is not be worth it.

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    /// map.insert("x-host", "localhost".parse().unwrap());
+    ///
+    /// assert_eq!(2, map.keys_len());
+    ///
+    /// map.insert("x-mime", "text/html".parse().unwrap());
+    ///
+    /// assert_eq!(2, map.keys_len());
+    /// ```
+    pub fn keys_len(&self) -> usize {
+        self.headers.len()
```
carllerche: Bug?

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    /// assert_eq!(map.get("x-host").unwrap(), &"hello");
+    /// assert_eq!(map.get("x-host").unwrap(), &"hello");
+    ///
+    /// map.append("x-host", "world".parse().unwrap());
+    /// assert_eq!(map.get("x-host").unwrap(), &"hello");
+    /// ```
+    pub fn get<K>(&self, key: K) -> Option<&MetadataValue>
+        where K: AsMetadataKey
+    {
+
```
carllerche: extra line

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    ///
+    /// for value in map.values() {
+    ///     println!("{:?}", value);
+    /// }
+    /// ```
+    pub fn values(&self) -> Values<MetadataValue> {
+        Values { inner: self.headers.values() }
+    }
+
+/*
```
carllerche: Could you remove this block if it is not needed?

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    ///     *counter = format!("{}{}", counter.to_str().unwrap(), "1").parse().unwrap();
+    /// }
+    ///
+    /// assert_eq!(map.get("content-length").unwrap(), "11");
+    /// assert_eq!(map.get("x-hello").unwrap(), "1");
+    /// ```
+    pub fn entry<K>(&mut self, key: K) -> Result<Entry<MetadataValue>, InvalidMetadataKey>
+        where K: AsMetadataKey
+    {
+        match key.entry(self) {
```
carllerche: Can this be implemented in terms of `HeaderMap::entry`?

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    /// assert!(map.insert("x-host", "world".parse().unwrap()).is_none());
+    /// assert!(!map.is_empty());
+    ///
+    /// let mut prev = map.insert("x-host", "earth".parse().unwrap()).unwrap();
+    /// assert_eq!("world", prev);
+    /// ```
+    pub fn insert<K>(&mut self, key: K, val: MetadataValue) -> Option<MetadataValue>
+        where K: IntoMetadataKey
+    {
+        key.insert(self, val)
```
carllerche: Can this be implemented in terms of HeaderMap::insert

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    ///
+    /// let values = map.get_all("x-host");
+    /// let mut i = values.iter();
+    /// assert_eq!("world", *i.next().unwrap());
+    /// assert_eq!("earth", *i.next().unwrap());
+    /// ```
+    pub fn append<K>(&mut self, key: K, value: MetadataValue) -> bool
+        where K: IntoMetadataKey
+    {
+        key.append(self, value)
```
carllerche: Can this be implemented in terms of HeaderMap::append?

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    ///
+    /// let prev = map.remove("x-host").unwrap();
+    /// assert_eq!("hello.world", prev);
+    ///
+    /// assert!(map.remove("x-host").is_none());
+    /// ```
+    pub fn remove<K>(&mut self, key: K) -> Option<MetadataValue>
+        where K: AsMetadataKey
+    {
+        key.remove(self)
```
carllerche: Can this be implemented in terms of HeaderMap::remove?

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    }
+}
+
+// ===== impl Iter =====
+
+impl<'a, T> Iterator for Iter<'a, T> {
+    type Item = (&'a str, &'a T);
```
carllerche: Should this iterate of `&MetadataKey`?

src/metadata_map.rs
```diff
@@ -0,0 +1,1716 @@
...
+    }
+}
+
+unsafe impl<'a, T: Sync> Sync for Drain<'a, T> {}
+unsafe impl<'a, T: Send> Send for Drain<'a, T> {}
+
+// ===== impl Keys =====
+
+impl<'a, T> Iterator for Keys<'a, T> {
+    type Item = &'a str;
```
carllerche: Should this iterate over `&MetadataKey`?

src/metadata_map.rs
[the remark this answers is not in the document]
per-gron: Ok, I'll keep them public then.

src/metadata_key.rs
per-gron: Ok, will remove.

src/metadata_map.rs
per-gron: Done.

src/metadata_map.rs
per-gron: Done.

src/metadata_map.rs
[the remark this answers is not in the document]
per-gron: Cool idea! Done.

src/metadata_map.rs
per-gron: Instead of removing this block, would it be possible for you to make a new release of hyperium/http? Then it would be possible to implement this method.

src/metadata_map.rs
per-gron: I'm too much of a Rust noob to be able to answer that question unfortunately. I want users of this function to be able to pass in `MetadataKey`s to this method. But `http::header::AsHeaderName` is a sealed trait so as far as I know I can't just add support for `MetadataKey` to it and replace `key.entry(self)` here with `self.headers.entry(key)` and make things work from there.

What did you have in mind?

src/metadata_map.rs
per-gron: See above

per-gron: See above

src/metadata_map.rs
per-gron: See above

carllerche:
@per-gron I pushed an `http` release: https://crates.io/crates/http/0.1.14

src/metadata_value.rs
```diff
@@ -0,0 +1,628 @@
...
+    /// assert!(!val.is_sensitive());
+    /// ```
+    #[inline]
+    pub fn is_sensitive(&self) -> bool {
+        self.inner.is_sensitive()
+    }
+
+    #[inline]
+    pub(crate) fn from_header_value(header_value: &HeaderValue) -> &Self {
+        unsafe { &*(header_value as *const HeaderValue as *const Self) }
```
seanmonstar: I believe the only way this isn't UB is if `MetadataValue` has the `#[repr(transparent)]` attribute on it. Otherwise, struct layout is undefined.

carllerche: As long as the struct has a single field and has no other repr on it, it is pretty much guaranteed to have the same layout. I don’t think it has been expressly stated, but std assumes this and so do many other crates. 

This technique is basically the only way to define your own DST and has been shown to me by rust team members. 

I think we can rely on it.

per-gron:
I fixed the last issues. I think this is ready to merge now.

src/metadata_value.rs
carllerche: I did it realize transparent was stable... that is the best option.

per-gron:
I'll go ahead and merge this now.

PR #95, opened 2018-11-24
Expose extensions map in Request

### LINKED ISSUE #94: Expose extensions map in Request
The underlying `http::Request` type from which the `tower_grpc::Request` type is made from has a useful method called [`extensions_mut()`](https://docs.rs/http/0.1.14/http/request/struct.Request.html#method.extensions_mut), which can store arbitrary data alongside the request, as long as it is `Send + Sync + 'static`. The data can then be queried later with:

```rust
if let Some(extra_metadata) = request.extensions().get::<MyMetadata>() {
    // Respond to request one way.
} else {
    // Respond to request another way.
}
```

It would be nice to preserve this metadata in the `tower_grpc::Request` and make it accessible through an `extensions()` method.

ebkalderon:
### Added

* Add `extensions()` and `extensions_mut()` methods to `Request`.

This is handy for storing extra metadata alongside a `Request` which can be processed in the server implementation. Closes #94.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -3,6 +3,7 @@ use http;
 #[derive(Debug)]
 pub struct Request<T> {
     headers: http::HeaderMap,
+    extensions: http::Extensions,
     message: T,
 }
 
@@ -11,6 +12,7 @@ impl<T> Request<T> {
     pub fn new(message: T) -> Self {
         Request {
             headers: http::HeaderMap::new(),
+            extensions: http::Extensions::new(),
             message,
         }
     }
@@ -35,6 +37,16 @@ impl<T> Request<T> {
         &mut self.headers
     }
 
+    /// Get a reference to the request extensions.
+    pub fn extensions(&self) -> &http::Extensions {
+        &self.extensions
+    }
+
+    /// Get a mutable reference to the request extensions.
+    pub fn extensions_mut(&mut self) -> &mut http::Extensions {
+        &mut self.extensions
+    }
+
     /// Consumes `self`, returning the message
     pub fn into_inner(self) -> T {
         self.message
@@ -45,6 +57,7 @@ impl<T> Request<T> {
         let (head, message) = http.into_parts();
         Request {
             headers: head.headers,
+            extensions: head.extensions,
             message,
         }
     }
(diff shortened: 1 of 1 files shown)


per-gron:
Hi Eyal! Thanks for your PR.

Instead of exposing `extensions()` and `extensions_mut()`, would it be possible for you to use gRPC [custom metadata](https://grpc.io/docs/guides/concepts.html#metadata)? They are essentially HTTP headers but with support for storing binary data too (gRPC libraries are supposed to automatically base64 encode/decode metadata fields whose names end with "-bin"). tower-grpc does not support binary custom metadata fields yet but I'm working on that in #92 and #90. There is some code in the [interop test suite](https://github.com/tower-rs/tower-grpc/blob/master/tower-grpc-interop/src/client.rs#L391) that shows how to set HTTP headers/custom metadata now, but that API is about to change a little bit with the changes I mentioned.

The reason I'm asking this is that I'm working on hiding HTTP from users of tower-grpc; In gRPC, HTTP is supposed to be an implementation detail, and it is designed so that gRPC interfaces should be language agnostic. For those reasons I think it's better to use gRPC's mechanism for this instead of further exposing HTTP and Rust specific things.

ebkalderon:
That sounds like a good idea. I was just looking at #92 and found your implementation interesting. Perhaps I could add comparable `metadata()` and `metadata_mut()` methods to `Request` and `Response`, and create a `MetadataMap` structured similarly to your `HeaderMap` proposed in that PR? If you'd like me to take that on, I might have a little time to work on it.

ebkalderon:
That sounds like a good idea. I was just looking at #92 and found your implementation interesting. I noticed you had a little bit of code for `MetadataMap` already present in there, but without any support for binary fields yet. Any thoughts on how you plan on getting that to work?

per-gron:
@ebkalderon Towards the end of the discussion in #90 is some information about the plans for adding support for binary fields. The tl;dr of it is that I am [working on it](https://github.com/per-gron/tower-grpc/blob/request-metadata-bin/src/metadata_map.rs) and expect to be able to upload a PR for that soon after #92 is merged and @carllerche has thought about what he wants the API to look like.

ebkalderon:
Sounds good. I think I will close this PR in the meantime as I wait for `MetadataMap` support to land. Thank you for working on this!

PR #97, opened 2018-11-26
Fix missing dyn in tower-grpc-build to satisfy rust 2018 idioms

LucioFranco:
When I compile my code using `tower-grpc-build` with these rust 2018 attributes:

```rust
#![warn(rust_2018_compatibility)]
#![warn(rust_2018_idioms)]
```

it produces a warning about not explicitly using `dyn`.

```bash
warning: trait objects without an explicit `dyn` are deprecated
   --> /Users/lucio/code/clique/target/debug/build/clique-a263e5c7d647a25f/out/clique.proto.rs:138:14
    |
138 |             <tower::Service<http::Request<grpc::BoxBody>, Response=_, Error=_, Future=_>>::poll_ready(self)
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn tower::Service<http::Request<grpc::BoxBody>, Response=_, Error=_, Future=_>`
    |
note: lint level defined here
   --> src/lib.rs:6:9
    |
6   | #![warn(rust_2018_idioms)]
    |         ^^^^^^^^^^^^^^^^
    = note: #[warn(bare_trait_objects)] implied by #[warn(rust_2018_idioms)]

warning: trait objects without an explicit `dyn` are deprecated
   --> /Users/lucio/code/clique/target/debug/build/clique-a263e5c7d647a25f/out/clique.proto.rs:143:14
    |
143 |             <tower::Service<http::Request<grpc::BoxBody>, Response=_, Error=_, Future=_>>::call(self, request)
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn tower::Service<http::Request<grpc::BoxBody>, Response=_, Error=_, Future=_>`
```

This change fixes this warning.

### DIFF
--- a/tower-grpc-build/src/server.rs
+++ b/tower-grpc-build/src/server.rs
@@ -292,7 +292,7 @@ macro_rules! try_ready {
             imp.new_fn("poll_ready")
                 .arg_mut_self()
                 .ret("futures::Poll<(), Self::Error>")
-                .line("<tower::Service<http::Request<grpc::BoxBody>, Response=_, Error=_, Future=_>>::poll_ready(self)")
+                .line("tower::Service::<http::Request<grpc::BoxBody>>::poll_ready(self)")
                 ;
 
             imp.new_fn("call")
@@ -300,7 +300,7 @@ macro_rules! try_ready {
                 .arg("request", "http::Request<tower_h2::RecvBody>")
                 .ret("Self::Future")
                 .line("let request = request.map(|b| grpc::BoxBody::new(Box::new(b)));")
-                .line("<tower::Service<http::Request<grpc::BoxBody>, Response=_, Error=_, Future=_>>::call(self, request)")
+                .line("tower::Service::<http::Request<grpc::BoxBody>>::call(self, request)")
                 ;
         }
     }


seanmonstar:
Thanks for trying to fix this! This is actually surprising, I didn't mean for it to try to do a dynamic dispatch call, I was trying to use UFCS to prevent conflicts that might occur with `self.poll_ready()`. I wonder why it wants to use dynamic dispatch. Is it because of the `<..>` around the whole thing?

seanmonstar:
I see, I just fiddled around with things in https://play.rust-lang.org. The `<tower::Service<http::Request<grpc::BoxBody>, Response=_, Error=_, Future=_>>::poll_ready(self)` make it choose dynamic dispatch. It should be `tower::Service::<http::Request<grpc::BoxBody>>::poll_ready(self)`.

Want to make that change and we can merge in?

LucioFranco:
@seanmonstar Good find! This seems like a way better solution. I've verified it works with my project that uses `tower-grpc-build`. 👍

seanmonstar:
Thank you!

PR #98, opened 2018-11-26
update examples from tokio-core to tokio

seanmonstar:


seanmonstar:
@per-gron yep! https://github.com/tower-rs/tower-grpc/blob/master/.travis.yml#L17

PR #102, opened 2018-12-05
Request metadata bin

### LINKED ISSUE #79: Metadata support

per-gron:
Adds support for base64 encoded "-bin" gRPC custom metadata.

Fixes #79.

Necessary for #27. See discussion at #90. This is a follow-up of #92.

Things to look for when reviewing:

* How do you like what the user-facing API ends up like?
* The `ValueEncoding` trait is afaik not importable from outside of this crate. Is that enough to prevent others from making implementations of it? It's not supposed to be possible.

Ping @carllerche here is a more or less fully fleshed out PR of what we discussed before.

### DIFF
--- a/src/metadata_encoding.rs
+++ b/src/metadata_encoding.rs
@@ -1,9 +1,16 @@
+use bytes::Bytes;
 use std::hash::Hash;
 
 // TODO(pgron): Make sealed
 pub trait ValueEncoding: Clone + Eq + PartialEq + Hash {
     #[doc(hidden)]
     fn is_valid_key(key: &str) -> bool;
+
+    #[doc(hidden)]
+    fn is_empty(value: &[u8]) -> bool;
+
+    #[doc(hidden)]
+    fn encode(value: Bytes) -> Bytes;
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
@@ -15,10 +22,28 @@ impl ValueEncoding for Ascii {
     fn is_valid_key(key: &str) -> bool {
         !Binary::is_valid_key(key)
     }
+
+    fn is_empty(value: &[u8]) -> bool {
+        value.is_empty()
+    }
+
+    fn encode(value: Bytes) -> Bytes {
+       value
+    }
 }
 
 impl ValueEncoding for Binary {
     fn is_valid_key(key: &str) -> bool {
         key.ends_with("-bin")
+    }
+
+    fn is_empty(value: &[u8]) -> bool {
+        // TODO(pgron): Do this properly for base64
+        value.is_empty()
+    }
+
+    fn encode(value: Bytes) -> Bytes {
+        // TODO(pgron): Do this properly for base64
+        value
     }
 }
--- a/src/metadata_encoding.rs
+++ b/src/metadata_encoding.rs
@@ -20,10 +20,19 @@ pub trait ValueEncoding: Clone + Eq + PartialEq + Hash {
     fn is_empty(value: &[u8]) -> bool;
 
     #[doc(hidden)]
+    fn from_bytes(value: &[u8]) -> Result<HeaderValue, InvalidMetadataValueBytes>;
+
+    #[doc(hidden)]
     fn from_shared(value: Bytes) -> Result<HeaderValue, InvalidMetadataValueBytes>;
 
     #[doc(hidden)]
     fn decode(value: &[u8]) -> Result<Bytes, InvalidMetadataValueBytes>;
+
+    #[doc(hidden)]
+    fn equals(a: &HeaderValue, b: &[u8]) -> bool;
+
+    #[doc(hidden)]
+    fn fmt(value: &HeaderValue, f: &mut fmt::Formatter) -> fmt::Result;
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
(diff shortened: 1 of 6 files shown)


src/metadata_value.rs
```diff
@@ -14,39 +20,33 @@ use metadata_key::MetadataKey;
...
-/// slice.
-#[derive(Debug)]
-pub struct InvalidMetadataValueBytes(InvalidMetadataValue);
-
 /// A possible error when converting a `MetadataValue` to a string representation.
 ///
 /// Metadata field values may contain opaque bytes, in which case it is not
 /// possible to represent the value as a string.
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
seanmonstar: I'd recommend removing the `Default`, since it a) allows creating a `ToStrError` outside the crate, and b) can possibly interfere if needing to add a field later that does not implement `Default`.

src/metadata_encoding.rs
```diff
@@ -0,0 +1,203 @@
+use bytes::Bytes;
+use http::header::HeaderValue;
+use std::fmt;
+use std::error::Error;
+use std::hash::Hash;
+
+/// A possible error when converting a `MetadataValue` from a string or byte
+/// slice.
+#[derive(Debug, Default)]
```
seanmonstar: I'd also remove the `Default` from this error.

src/metadata_encoding.rs
```diff
@@ -0,0 +1,203 @@
...
+pub struct InvalidMetadataValue {
+    _priv: (),
+}
+
+pub trait ValueEncoding: Clone + Eq + PartialEq + Hash {
```
seanmonstar: Is this meant to be a "sealed" internal trait? As in, only types within this crate implement it, and users of the crate don't actually ever touch the trait? If so, there's some Rust type system trickery that we can do to make the trait less "public":

- Add an internal private module `sealed`, with a `pub trait Sealed` inside it.
- Add `self::sealed::Sealed` as a super trait of `ValueEncoding`.
- Implement `Sealed` for `Ascii` and `Binary`.
- Optionally move all the `#[doc(hidden)]` methods to the `Sealed` trait, so no one can even call them. This allows changing their signature without breaking external code.

src/metadata_encoding.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    #[doc(hidden)]
+    fn values_equal(a: &HeaderValue, b: &HeaderValue) -> bool;
+
+    #[doc(hidden)]
+    fn fmt(value: &HeaderValue, f: &mut fmt::Formatter) -> fmt::Result;
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct Ascii {}
```
seanmonstar: If these are just marker types, and should never be instantiated, consider making them empty enums.

src/metadata_encoding.rs
```diff
@@ -0,0 +1,203 @@
...
+        } else {
+            write!(f, "b[invalid]{:?}", value)
+        }
+    }
+}
+
+// ===== impl InvalidMetadataValue =====
+
+impl InvalidMetadataValue {
+    pub fn new() -> Self {
```
seanmonstar: I'd make error constructors `pub(crate)`.

src/metadata_encoding.rs
per-gron: Thanks for the info on how to do this, I will update the PR when I get time.

src/metadata_encoding.rs
per-gron: Nice idea, will do.

src/metadata_value.rs
per-gron: (This comment is a response to all of your comments in this review related to this.)

I'm open to changing my mind here, especially since I'm a Rust newbie, but to me hiding constructors of error types seems like a bad idea. I can't recall seeing a design in some other language that attempts to do this.

Conceptually, I think of error and exception types as value types (as in [value vs identity](https://www.youtube.com/watch?v=-6BsiVyC1kM)), and, being value types, I do not see any issue in allowing anyone to create any kind of error value they desire.

Practically I care about this because of testing. Disallowing any other code to create these error values requires any code that uses this to immediately convert the error to some other type, even thin wrappers that don't do much, otherwise it is impossible to test error conditions.

It's true that it becomes impossible to add mandatory non-defaultable fields without an API break but the probability of that happening, combined with the small pain of being forced to use `Option` or similar, vs the relatively large testability pain makes it seem like a bad tradeoff to not allow tests to create empty error values.

hawkw: My $0.02 is to agree with @seanmonstar on this. Although I definitely understand @per-gron's point here, in my experience, it's pretty important for libraries to avoid anything that has the potential to cause a breaking change, even if it doesn't seem particularly likely. And, as @seanmonstar points out, this error is relatively easy to generate in testing situations.

per-gron: Thanks for taking the time to explain more thoroughly. It still seems a bit odd to me (especially given that the worst case scenario is to have to use `Option`) but I see your point about being conservative. I've updated the PR to make the constructors `pub(crate)` and remove `Default`.

src/metadata_encoding.rs
per-gron: Done.

per-gron:
@carllerche do you also want to review this PR before I merge? I'll probably merge this tomorrow unless I get more comments.

carllerche:
I'm back now! I will try to take a look in the next 24 hours. I'm still catching up on email :)

per-gron:
@carllerche Aha, I see! No worries, I'm in no particular rush with this really. I'm off to vacation myself next week and will likely not spend much time on this until January.

PR #105, opened 2018-12-10
require http 0.1.14, change MakeService to tower_util import

seanmonstar:


per-gron:
I get [CI failures](https://travis-ci.org/tower-rs/tower-grpc/jobs/466445639) in #102 that seem like they would be fixed by this so I'll go ahead and merge this now.

PR #107, opened 2018-12-27
Add clone impl for clients when the inner service impls clone

LucioFranco:
This change adds the ability to clone the outter client when the inner service is cloneable. This allows one to embed tower-buffer for example into the inner service to provide a cloneable and buffer backed client.

```rust
    let say_hello = make_client.make_service(())
        .map(move |conn| {
            use hello_world::client::Greeter;
            use tower_http::add_origin;

            let conn = add_origin::Builder::new()
                .uri(uri)
                .build(conn)
                .unwrap();

			let conn = Buffer::new(conn, 100).unwrap();

            Greeter::new(conn)
        })
        .and_then(|client| {
            use hello_world::HelloRequest;

			let mut client = client.clone();

            client.say_hello(Request::new(HelloRequest {
                name: "What is in a name?".to_string(),
            })).map_err(|e| panic!("gRPC request failed; err={:?}", e))
        })
```

This shoud also enable the ability to embed a client into a server for use when calling out to other grpc/tower clients.

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -10,7 +10,7 @@ use tower_http::HttpService;
 
 use body::{Body, BoxBody};
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Grpc<T> {
     /// The inner HTTP/2.0 service.
     inner: T,
--- a/tower-grpc-build/src/client.rs
+++ b/tower-grpc-build/src/client.rs
@@ -51,6 +51,7 @@ impl ServiceGenerator {
             .vis("pub")
             .generic("T")
             .derive("Debug")
+            .derive("Clone")
             .field("inner", "grpc::Grpc<T>")
             ;
     }


LucioFranco:
I figured that derive only implemented it if the inner items of the struct did. So this would enforce that the inner service is cloneable or it would error out. Which I don’t think is very flexible then only implementing it if the inner derive is clone.

hntd187:
Couldn't you just mark both with derive(Clone) instead? the inner isn't anything else except Grpc right?

seanmonstar:
When a type has a generic, the derive generates the same code you wrote. The benefit is we don't need to keep the manual impl in sync if other fields are added.

LucioFranco:
@seanmonstar though wouldn’t adding derive(Clone) add a T: Clone bound to the struct? I’m not too sure I can check a bit later when I get back.

seanmonstar:
Nope, derive doesn't require the trait be implemented, just provides an impl if the generic does implement the trait.

LucioFranco:
@seanmonstar I've updated it to now just use `derive(Clone)` thanks for letting me know about that! I honestly, looked at doing it that way but the documentation on it wasn't clear (or I wasn't looking in the right place). Found it explained in an issue that pretty much this avoids breaking a bunch of apis.

PR #109, opened 2019-01-09
Derive Clone impl for Error

mzabaluev:
It's sometimes convenient if the whole error value is cloneable (like `Status` is), provided that its `Inner` variant's content is cloneable, which it is by default.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -2,7 +2,7 @@ use h2;
 use std;
 use std::fmt;
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


carllerche:
Errors tend to not be `Clone`. Could you elaborate more on the use case?

LucioFranco:
@mzabaluev yeah, I agree with @carllerche on this one, I would rather not add the clone bound as you said its most likely to work around the compiler. If you want you can hop into gitter and we can look at the error types and try to get something to work 😄. 

Gonna close this, feel free to reopen if you have any more thoughts/changes

PR #110, opened 2019-01-12
interop: bump rustls to 0.14.0 to make it compile

DebugSteven:
Several past versions of `untrusted`, a dependency of `rustls`,
were recently yanked from crates.io. This broke the version of 
`rustls` we previously depended on for the interop tests.

This branch updates the rustls version to one that depends on
a version of `untrusted` that was not yanked.


hawkw:
Just for my understanding, what's the error that occurs on a clean build without bumping rustls?

DebugSteven:
thank you!

hawkw:
...and merged! Thanks for fixing the build!

Issue #96, opened 2018-11-24
Retain inline Protobuf comments in generated code


ebkalderon:
One nice feature of Prost is its ability to retain the inline comments on Protobuf types and their fields as doc comments in the generated Rust code. However, it seems that the builder from `tower-grpc-build` doesn't retain them for the client/server traits, structs, and methods. It would be nice to include those comments in the generated code as well.

hawkw:
Fixed by #112!

ebkalderon:
@hawkw Fantastic!

PR #112, opened 2019-02-05
Retain protobuf comments

DebugSteven:
grpc-build: generated code comments for methods, structs, and traits

Currently, the Rust code generated from protobuf files with comments
doesn't include the comments.

This branch fixes this by including the comments for
methods and structs on the client side and methods and traits on the
server side. This is done by calling doc on the sections of code that
generate new functions, structs, and traits. In the proto file, the
service comments are used for the structs and traits and the functions
comments inside those services are for the comments on methods. This
handles multiline comments that start immediately before functions
and services.

Fixes: #96

### DIFF
--- a/tower-grpc-build/src/client.rs
+++ b/tower-grpc-build/src/client.rs
@@ -1,4 +1,5 @@
 use codegen;
+use comments_to_rustdoc;
 use prost_build;
 use super::ImportType;
 
@@ -52,11 +53,7 @@ impl ServiceGenerator {
             .generic("T")
             .derive("Debug")
             .derive("Clone")
-            .doc(&service.comments.leading
-                 .iter()
-                 .fold(String::new(), |acc, s|{
-                     acc + s.trim_start() + "\n"
-                 }))
+            .doc(&comments_to_rustdoc(&service.comments.leading))
             .field("inner", "grpc::Grpc<T>")
             ;
     }
@@ -100,11 +97,7 @@ impl ServiceGenerator {
                 .bound("T::ResponseBody", "grpc::Body")
                 .arg_mut_self()
                 .line(format!("let path = http::PathAndQuery::from_static({});", path))
-                .doc(&method.comments.leading
-                     .iter()
-                     .fold(String::new(), |acc, s|{
-                         acc + s.trim_start() + "\n"
-                     }))
+                .doc(&comments_to_rustdoc(&service.comments.leading))
                 ;
 
             let mut request = codegen::Type::new("grpc::Request");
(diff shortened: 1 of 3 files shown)


tower-grpc-build/src/client.rs
```diff
@@ -52,6 +52,11 @@ impl ServiceGenerator {
             .generic("T")
             .derive("Debug")
             .derive("Clone")
+            .doc(&service.comments.leading
```
hawkw: I _think_ using only the `leading` field on `service.comments` here (and elsewhere in this branch) is probably the right thing, as I'm guessing the `leading_detached` comments (which are separated from the service declaration by one or more newlines) are not typically describing the API of the service definition. 

However, if anyone else thinks we ought to include both sets of comments (@ebkalderon?) we can always add them in a subsequent PR!

tower-grpc-build/src/client.rs
```diff
@@ -95,6 +100,11 @@ impl ServiceGenerator {
                 .bound("T::ResponseBody", "grpc::Body")
                 .arg_mut_self()
                 .line(format!("let path = http::PathAndQuery::from_static({});", path))
+                .doc(&method.comments.leading
```
hawkw: Take it or leave it, but it _might_ be worth factoring out the repeated code into a helper function? That way, we can just say things like
```rust
// ...
.doc(comments_to_rustdoc(&method.comments))
// ...
```
or something like that, here and elsewhere?

This would also have the advantage that if we were to change this logic later (such as to add the `leading_detached` comments, for example) we would only need to make the change in one place.

DebugSteven: I'm happy to make this change if you would like to wait to merge it. I appreciate the feedback!

hawkw: It's up to you! We can also merge this now and make that change in a follow-up PR, if you prefer. 

Just let me know if you're planning on addressing it in this branch; otherwise I'll go ahead and merge the PR.

DebugSteven: I'll address it in this branch. I'll push up those changes soon.

tower-grpc-build/src/lib.rs
```diff
@@ -226,8 +226,9 @@ fn to_upper_camel(s: &str) -> String {
 }
 
 /// This function takes an array of comments and properly formats them.
```
hawkw: Tiny nit: I think this comment ought to be updated now that it takes a `&prost::Comments`.

Also, I _might_ rephrase  "properly formats them" as "formats them as a `String` to be used as a RustDoc comment", or something?

hawkw:
@DebugSteven this looks great, thanks for putting up with all my nit-picking!

PR #116, opened 2019-02-27
Remove the `Error` type in favor of only `Status`

seanmonstar:
The `Inner` variant on the `Error` trait caused confusion, and was the
source of creating some vague `Status`s. This removes that enum
entirely, requiring that people create an `Status` instead of
`Error::Inner(())` and the library being unsure of what exactly that
means.


seanmonstar:
@per-gron I remember you wanted this done, but felt like there was something preventing us from doing so. I don't recall what that may have been, do you?

src/generic/codec.rs
```diff
@@ -82,6 +81,12 @@ enum EncodeInner<T, U> {
     Err(Status),
 }
 
+#[derive(Debug)]
+enum Role {
```
hawkw: 👍

seanmonstar:
>  should `Status` maybe now be renamed to `Error` or `ErrorStatus`?

Hm, maybe! It looks like grpc-go uses `status.Error`. I haven't look at the other language impls...

per-gron:
> @per-gron I remember you wanted this done, but felt like there was something preventing us from doing so. I don't recall what that may have been, do you?

I'm not sure I remember, but maybe it was this: Tower's design is based on stacking middleware, and part of how that is implemented in practice is that the error types of middleware are nested using enums with "inner". So if we want to expose the `Status` enum as the only error type visible to `tower-grpc` users, then there will have to be something that converts errors from other middleware into the `tower-grpc` `Status` type.

per-gron:
> > should `Status` maybe now be renamed to `Error` or `ErrorStatus`?
> 
> Hm, maybe! It looks like grpc-go uses `status.Error`. I haven't look at the other language impls...

The C++ implementation calls it `Status`: https://github.com/grpc/grpc/blob/master/include/grpcpp/impl/codegen/status.h

C# calls it `Status`: https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/Status.cs

Java calls it `Status`: https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/Status.java

src/client/mod.rs
```diff
@@ -31,10 +31,14 @@ impl<T> Grpc<T> {
...
-    pub fn poll_ready<R>(&mut self) -> Poll<(), ::Error<T::Error>>
-    where T: HttpService<R> {
+    pub fn poll_ready<R>(&mut self) -> Poll<(), ::Status>
+    where
+        T: HttpService<R>,
+    {
         self.inner.poll_ready()
-            .map_err(::Error::Inner)
+            .map_err(|_| {
+                ::Status::unknown("inner client service poll_ready error".into())
```
per-gron: I think this is the thing that made me think doing this is difficult; here you are throwing away error information, no?

src/client/streaming.rs
```diff
@@ -31,17 +31,16 @@ where T: Message + Default,
...
         use codec::Decoder;
         use generic::Streaming;
 
-        let response = self.inner.poll()
-            .map_err(::Error::Inner);
-
         // Get the response
-        let response = try_ready!(response);
+        let response = try_ready!(self.inner.poll().map_err(|_| {
+            ::Status::unknown("inner response future error".into())
```
per-gron: also here

tower-grpc-interop/src/client.rs
```diff
@@ -641,14 +640,13 @@ impl TestClients {
     fn unimplemented_method_test(&mut self)
             -> impl Future<Item=Vec<TestAssertion>, Error=Box<Error>> {
```
per-gron: This could be done later, but I think it is possible that this PR removes the need for `Box<Error>`; the reason that is used is because there were multiple error types because of the generic `Error`.

src/client/mod.rs
seanmonstar: That's true. I tried requiring `HttpService<Error = impl Into<Status>>`, but it was irritating with the examples and interop.

I'm going to try just requiring `Error: std::error::Error`, since that can be used to downcast to `Status`, or otherwise make an `UNKNOWN` status with the error as the message.

src/client/mod.rs
```diff
@@ -31,10 +33,15 @@ impl<T> Grpc<T> {
         Grpc { inner }
     }
 
-    pub fn poll_ready<R>(&mut self) -> Poll<(), ::Error<T::Error>>
-    where T: HttpService<R> {
+    pub fn poll_ready<R>(&mut self) -> Poll<(), ::Status>
+    where
+        T: HttpService<R>,
+        T::Error: Error + 'static,
```
carllerche: This bound will not work with any service that has an error of type `Box<Error>` I believe.

carllerche:
Thanks for getting this done 👍 Looks great over all. I just don't know how `T::Error: Error + 'static` will play with https://github.com/tower-rs/tower/issues/131

PR #119, opened 2019-03-01
Refine the `Status` type and error bounds

seanmonstar:
- Adds `Status::new` to replace `with_code` and `with_code_and_message`.
- Renames `error_message()` to `message()`, and `binary_error_details`
  to `details`.
- Changes `E: Error + 'static` bounds on services to `Into<Box<dyn Error>>`
  bounds, since `Box<dyn Error>` doesn't satisfy `E: Error`.


seanmonstar:
Ya, we could adjust `from_error` to not clone from the boxed Status if it is one, but once the issue is fixed in libstd, I think we'd revert back.

PR #120, opened 2019-03-02
Allow Clients to pass Request<M> or M

seanmonstar:
I can't decide if this is a good idea or not. At first thought, it sounds nice, since it allows users to skip importing and creating a `grpc::Request` if they don't need to set any metadata...

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -1,6 +1,7 @@
 use http;
 use metadata::MetadataMap;
 
+/// A gRPC request containing an message and metadata.
 #[derive(Debug)]
 pub struct Request<T> {
     metadata: MetadataMap,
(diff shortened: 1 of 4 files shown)


carllerche:
I'm +1 on this idea.

LucioFranco:
so if I understand correctly, you can still pass a request but if you pass the request item it will just wrap it into `Request::new`. I'm +1 on this QOL improvement.

seanmonstar:
It seems inference is unhappy in the client-streaming case. Maybe it should only be if the client is sending a single message?

carllerche:
@seanmonstar how did the client case compile before? I'm not seeing how the change would impact inference...

carllerche:
Ideally both streaming and single message would operate the same way...

PR #121, opened 2019-03-02
Add metadata to Response

seanmonstar:


per-gron:
Thanks for doing this! I meant to do this but never got to it.

PR #122, opened 2019-03-03
Impl for hyper::Body

hntd187:
This adds a feature for an optional impl of `hyper::Body` I don't think this is 100% correct so I'm expecting to have to do some more work on this, but for the most part it at least compiles and generates compilable code. The impl gives a recursion which is probably possible to remove, but I'm not sure how to accomplish that.

Also, how is testing setup for the project? I'm not sure where tests for this should live.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,8 @@ authors = [
 license = "MIT"
 
 [features]
-default = ["protobuf", "tower-h2"]
+default = ["protobuf", "tower-h2", "hyper-body"]
+hyper-body = ["hyper"]
 protobuf = ["prost"]
 
 [workspace]
@@ -34,6 +35,7 @@ http = "0.1.14"
 h2 = "0.1.11"
 log = "0.4"
 percent-encoding = "1.0.1"
+hyper = { version = "0.12.25", optional = true }
 tower-h2 = { git = "https://github.com/tower-rs/tower-h2", optional = true }
 tower-http = { git = "https://github.com/tower-rs/tower-http" }
 tower-service = { git = "https://github.com/tower-rs/tower"  }
--- a/src/body.rs
+++ b/src/body.rs
@@ -1,8 +1,9 @@
 use std::fmt;
 
 use bytes::{Bytes, IntoBuf};
-use futures::Poll;
+use futures::{Async, Poll};
 use http;
+use Code;
 
 /// A body to send and receive gRPC messages.
 pub trait Body {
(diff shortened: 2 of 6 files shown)


src/body.rs
```diff
@@ -62,6 +63,47 @@ where
...
+#[allow(unconditional_recursion)]
+impl<T> ::hyper::body::Payload for BoxBody<T>
+where
+    T: bytes::Buf + Send + IntoBuf + 'static,
+{
+    type Data = T;
+    type Error = ::hyper::Error;
+
+    fn poll_data(&mut self) -> Poll<Option<Self::Data>, Self::Error> {
+        ::hyper::body::Payload::poll_data(self)
```
LucioFranco: this should call `Body::poll_data(self)` I think, this is a recusive call because you are implementing payload and then calling yourself. You want to call the `tower_grpc::body::Body` trait impl i think. Thats how tower-h2 does it.

src/body.rs
```diff
@@ -62,6 +63,47 @@ where
...
+}
+
+#[cfg(feature = "hyper-body")]
+impl Body for ::hyper::Body {
+    type Data = Bytes;
+
+    fn poll_data(&mut self) -> Poll<Option<Self::Data>, ::Status> {
+        match ::hyper::body::Payload::poll_data(self) {
+            Ok(Async::Ready(Some(chunk))) => Ok(Async::Ready(Some(chunk.into_bytes()))),
+            Ok(Async::Ready(None)) => unreachable!(),
```
LucioFranco: why is this unreachable?

Cargo.toml
```diff
@@ -9,7 +9,8 @@ authors = [
 license = "MIT"
 
 [features]
-default = ["protobuf", "tower-h2"]
+default = ["protobuf", "tower-h2", "hyper-body"]
+hyper-body = ["hyper"]
```
LucioFranco: I'd prob just call this `hyper` that aligns better with the other trait names.

src/body.rs
hntd187: I realized this is wrong, it should be `Ok(Async::Ready(None))`

Cargo.toml
hntd187: Can't because hyper is a dependency name

carllerche: All optional dependencies are automatically features, so you don't need to explicitly list out the feature.

carllerche:
In general, all of these types are being moved into `tower-http` (for example, see here: https://github.com/tower-rs/tower-h2/pull/54).

I'm not sure what @seanmonstar's timeline to add a tower dependency to hyper. I'm unsure if `tower-http-service` should add a `hyper` dep or we wait for hyper to do so.

Cargo.toml
hntd187: Ahh okay so I can just take it out entirely

carllerche: Yes, but see the comment below first.

hntd187:
Discussed this in gitter, gonna hold off for what Carl mentioned above.

PR #123, opened 2019-03-04
Fix server's generated unimplemented catch-all

seanmonstar:
Previously, since the generated error body would yield `true` for
`is_end_stream`, the `grpc-status` would never be sent in the trailers.

This adds some basic server interop tests which would see this bug.

### DIFF
--- a/tower-grpc-build/src/server.rs
+++ b/tower-grpc-build/src/server.rs
@@ -208,33 +208,33 @@ macro_rules! try_ready {
                                 "let service = {}::methods::{}(self.{}.clone());",
                                 lower_name, &upper_name, lower_name));
 
-                        handle.line("let response = grpc::Grpc::unary(service, request);");
+                        handle.line("let response = grpc::unary(service, request);");
                     }
                     (false, true) => {
                         handle.line(&format!(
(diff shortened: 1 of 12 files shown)


tower-grpc-build/src/server.rs
```diff
@@ -411,32 +412,38 @@ macro_rules! try_ready {
...
-
-        let mut poll_metadata_catch_all = codegen::Block::new("Err(ref status) =>");
-            poll_metadata_catch_all
-            .line("status.to_header_map().map(Some).map(Into::into)")
-            ;
-
-            poll_metadata_block.push_block(poll_metadata_catch_all);
+            is_end_stream_block
+                .line(&format!(
+                    "{}(_) => true,",
```
carllerche: I'm a bit confused. The PR description stated that the problem is that the unimplemented variant's `is_end_stream()` fn would return `true`, so trailers were not sent. However, as far as I can tell, this behavior is still present.

seanmonstar: Instead of just changing the `is_end_stream` function, so that it would be returned in the trailers, this changes it so that the status is returned in the headers.

PR #128, opened 2019-03-06
Replace Body trait with that from tower-http-service

seanmonstar:


seanmonstar:
Oh huh, the change to a standard `Body` trait means the adapter tower-grpc had for tower-h2 isnt used anymore, which was automatically releasing window capacity when converting `tower_h2::Data` into `Bytes`...

seanmonstar:
PR to tower-h2 to just release immediately: https://github.com/tower-rs/tower-h2/pull/57

PR #132, opened 2019-03-07
Remove tower-h2 feature, generate server services generic over request body

seanmonstar:
Previously, the generated server services wanted `http::Request<tower_grpc::BoxBody>`. In order to make the examples Just Work out of the box with tower-h2, there was an optional tower-h2 feature to produce a second `Service` impl for `tower_h2::RecvBody`.

Instead, this removes that feature, and generates a single generic `Service` that will automatically box up the body internally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ authors = [
 license = "MIT"
 
 [features]
-default = ["protobuf", "tower-h2"]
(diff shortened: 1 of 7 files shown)


seanmonstar:
Hm, trying this out in the proxy, it's now conflicting if I construct a `BoxBody` beforehand. I may back that part out...

PR #135, opened 2019-03-09
to allow encoding of Status including utf-8 char

Emanon42:


src/status.rs
```diff
@@ -129,7 +129,10 @@ impl Status {
         header_map.insert(GRPC_STATUS_HEADER_CODE, self.code.to_header_value());
 
         if !self.message.is_empty() {
-            header_map.insert(GRPC_STATUS_MESSAGE_HEADER, HeaderValue::from_str(&self.message)
+            header_map.insert(GRPC_STATUS_MESSAGE_HEADER, HeaderValue::from_str(
+                    Box::leak(utf8_percent_encode(&self.message(), DEFAULT_ENCODE_SET)
```
per-gron: Why use `Box::leak` here? That sounds like it will.. leak memory, no? Isn't it easier to do

    utf8_percent_encode(...).to_string().parse().unwrap()

instead of the `HeaderValue::from_str(...)` call (maybe also handle the potential failure there)

seanmonstar:
I've just added server support for `status_code_and_message` in https://github.com/tower-rs/tower-grpc/pull/134, so adding a test in this PR (once that one is merged) would just be to add `special_status_message` to list of test cases tested.

src/status.rs
seanmonstar: - To save on allocations, the owned string can be converted into a `Bytes` and then [`HeaderValue::from_shared`](https://docs.rs/http/0.1.16/http/header/struct.HeaderValue.html#method.from_shared) can be called.
- Looking at other grpc implementations, it looks like we want to define an encode set that encodes anything outside of the visible ASCII range.

Emanon42: Thanks for reviewing! About the encoding set for Status message, I browsed the grpc-java and grpc-go. Looks like they directly use percent encoding - the byte is converted into hexadecimal notation with a '%' prepended without custom encoding set.
But for some Unicode char like emoji, `percent_encoding::decode_utf8()` sometimes can't give correct parsing. E.g.
```
let s1 = Status::new(Code::Unknown, "Unicode non-BMP 🌶️").to_header_map();
let s2 = Status::from_header_map(&s2);
```
will give s2 as `Status { code: Unknown, message: "Unicode non-BMP 🌶\u{fe0f}" }` which `\u{fe0f}` should not appear.

Emanon42: Also for adding unit tests for `Status`, is that necessary to implement `PartialEq` for `Status` in order to use assert?

seanmonstar: No, there shouldn't be need to add `PartialEq for Status` (in fact, that's probably not a great idea in general).

Should be good enough to just use two asserts, `assert_eq!(status.code(), expected_code)` and `assert_eq!(status.message(), expected_msg)`.

seanmonstar:
I think it may also be beneficial to optimize for a fast path when the `message` doesn't need to be encoded. Something like checking all the bytes for `DEFAULT_ENCODE_SET.contains`, and if a byte *does* need to be encoded, only start from the "bad" byte.

This way, a message like "oh no it didn't work", which doesn't need to be encoded, doesn't result in allocating a wasted string.

Emanon42:
> I think it may also be beneficial to optimize for a fast path when the `message` doesn't need to be encoded. Something like checking all the bytes for `DEFAULT_ENCODE_SET.contains`, and if a byte _does_ need to be encoded, only start from the "bad" byte.
> 
> This way, a message like "oh no it didn't work", which doesn't need to be encoded, doesn't result in allocating a wasted string.

To encode from "bad" byte will break original `&[u8]` and allocates new data, which would cost more on array split and merge I guess? So currently I just add a simple quick path for the message which are all "good" bytes.

seanmonstar:
> So currently I just add a simple quick path for the message which are all "good" bytes.

That's fine. Note that if you import `[EncodeSet](https://docs.rs/percent-encoding/1.0.1/percent_encoding/trait.EncodeSet.html)`, you can use `DEFAULT_ENCODE_SET.contains` instead of duplicating the bytes range check.

Finally, the server-interop tests were merged to master, so we can add a test by adding `special_status_message` to https://github.com/tower-rs/tower-grpc/blob/master/tower-grpc-interop/travis-interop.sh#L14

src/status.rs
```diff
@@ -214,9 +214,23 @@ impl Status {
         header_map.insert(GRPC_STATUS_HEADER_CODE, self.code.to_header_value());
 
         if !self.message.is_empty() {
-            header_map.insert(GRPC_STATUS_MESSAGE_HEADER, HeaderValue::from_str(&self.message)
-                .map_err(invalid_header_value_byte)?);
+            let good_bytes = &self.message.as_bytes().iter().all(|&x| DEFAULT_ENCODE_SET.contains(x));
+            let to_write = 
+            match good_bytes{
```
seanmonstar: The name `good_bytes` might not give the best description. To me, this reads as "is the message all good_bytes? yes: percent encode them. no: send as-is". We'd want the opposite sentence.

Now, the tests appear to pass, so the *code* is right. We just want to fix up the narrative. I'd change the variable name to `needs_encode` or similar.

And then, since this is a boolean, it's more idiomatic in Rust to just use `if`, like `if needs_encoded { ... } else { ... }`.

src/status.rs
```diff
@@ -214,9 +214,23 @@ impl Status {
...
+                true => {
+                    Bytes::from(
+                        percent_encode(&self.message().as_bytes(), DEFAULT_ENCODE_SET)
+                        .to_string()
+                        .as_bytes())
```
seanmonstar: This is a slight gotcha: when percent-encoding, we have a new `String` (from `to_string()`), which owns some memory. It's nearly free to convert a `String` into a `Bytes`, as it just assumes ownership of that memory. By calling `Bytes::from(String.as_bytes())`, this is asking `Bytes` to create a new slot of memory and copy, since `as_bytes` returns a borrow of the `String`s memory.

So, this could just be `percent_encode(..).to_string().into()` (type inference will detect what `into` will convert to because the other branch is also returning `Bytes`.

Emanon42: Thanks! Learn a new way to use `into()`

Issue #138, opened 2019-03-11
Update to prost 0.5


seanmonstar:
Hopefully should be easy, [there doesn't seem to be much breaking changes](https://github.com/danburkert/prost/releases/tag/v0.5.0).

PR #139, opened 2019-03-12
Update prost to v0.5

csharad:
~The code does not break as its building fine.~ Successfully ran the `routeguide` example. 
It seems it does have changes:
> prost 0.4.0 escapes field and message names which are Rust keyword by appending a suffix underscore; prost 0.5.0 does not escape field or message names, instead using the new raw-identifier feature

#138


tower-grpc-build/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license = "MIT"
 
 [dependencies]
 codegen = { git = "https://github.com/carllerche/codegen" }
-prost-build = "0.4"
```
06kellyjac: This is the offending change. When I put it back to `"0.4"` everything else builds fine.
That's as far as I was able to get in the short amount of time I had. Might be able to take another look later.

tower-grpc-build/Cargo.toml
```diff
@@ -6,7 +6,7 @@ license = "MIT"
...
+prost-build = "0.5"
```
06kellyjac: This is the offending change. When I put it back to `"0.4"` everything else builds fine.
That's as far as I was able to get in the short amount of time I had. Might be able to take another look later.

06kellyjac: This is the offending change. When I put it back to `"0.4"` everything else builds fine.
That's as far as I was able to get in the short amount of time I had. Might be able to take another look later.

seanmonstar:
For importing `()`, could it be the exact string here has changed? https://github.com/tower-rs/tower-grpc/blob/master/tower-grpc-build/src/lib.rs#L193

csharad:
> could it be the exact string here has changed? 

No, the string is still the same.

I was able to pin-point the reason of one of the imports (within `mod client`).
https://github.com/tower-rs/tower-grpc/blob/94a7b51cd2e54a5bd8321a6075c00c118639abfc/tower-grpc-build/src/lib.rs#L129-L132

Here the `ty = "()"` and is getting imported. For other places I have not been able to find the source.

seanmonstar:
I've fixed up the failure trying to import `()` in https://github.com/tower-rs/tower-grpc/pull/140

csharad:
Cool! Closing this PR then.

PR #146, opened 2019-03-28
quick hack for include_proto!() proc macro

Emanon42:
That is first time for me to design a proc macro, please indicate the flaw in my code, thanks!

new `include_proto!()` macro receive following 4 parameters, split by comma :
- `.proto` file path
- proto folder path
- service name (as name of local mod)
- messages names (multiple names split by vertical bar `|`)

There is an issue in my implementation: I have to provide absolute path for proto files. Since proc macro is in another crate, it can not acquire `CARGO_MANIFEST_DIR` of example. Could anyone give me a hint to solve this? Thanks!


kingoftheconnors:
Do you know if the error with Travis is just because Travis has a phobia of no-parameter calls? If so, you could try using the path: `.` to refer to the current directory while still having something to fill the arguments.

Emanon42:
> Do you know if the error with Travis is just because Travis has a phobia of no-parameter calls? If so, you could try using the path: `.` to refer to the current directory while still having something to fill the arguments.

proc macro treats anything in its parentheses as `TokenStream`, so I guess using `.` could only refer to directory of `tower-grpc-build-procm` instead of `tower-grpc-build`

Emanon42:
Also should I PR in another branch? Do we have a `gsoc` branch?

kingoftheconnors:
I... don't think so? I'm just applying as well, so I have no idea, but a github search doesn't pull anything up and this repo doesn't have any gsoc branches. So I don't _think_ so.

kingoftheconnors:
Hey, it passed!

Emanon42:
> Hey, it passed!

No, I didn't solve the problem... user still need to type to give full absolute path to proc macro, I just copy and paste travis directory 😖

Emanon42:
I read `syn` crate document and find a better implementation:
```
struct ProtoParas {
    proto_dir: Ident,
    include_dir: Ident,
    service_name: Ident,
    msgs_name: Expr,
}

impl Parse for ProtoParas {
    fn parse(input: ParseStream) -> Result<Self> {
        // parse token stream
    }
}

#[proc_macro]
pub fn include_proto(input: TokenStream) -> TokenStream {
    let ProtoParas {
        proto_dir,
        include_dir,
        service_name,
        msgs_name,
    } = parse_macro_input!(input as ProtoParas);
    // do other jobs
```
With `syn::Parse` we can give custom warning message. But it still doesn't solve the issue of full absolute directory. Is there a method to acquire other cargo directory in proc macro cargo?

carllerche:
`cargo` sets a bunch of [env variables](https://doc.rust-lang.org/cargo/reference/environment-variables.html). Could one of those help avoid the absolute path?

.gitignore
```diff
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+.idea/
```
carllerche: Maybe add this file to your [global git ignore list](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c) instead?

carllerche:
Also, I don't think you can rely on `OUT_DIR` here. The cargo documentation says that this env var is only set if there is a `build.rs` file.

carllerche:
A more incremental first step could be to still require the `build.rs` and define `include_proto!` to look for files in `$OUT_DIR`.

Right now, the proto mod structure [must be created manually](https://github.com/linkerd/linkerd2-proxy-api/blob/master/rs/src/lib.rs#L21-L41). That entire block could be replaced by `include_proto!()` while still maintaining `build.rs` to generate the `.rs` files.

.gitignore
Emanon42: Sorry I wrongly commit this change. I have removed it.

Emanon42:
Got it. The end of semester is coming and exam is eating my time. I will update this weekend (as much as I can).

Emanon42:
Now it could receive multiple pairs of `.rs` file and local module name pair (wrapped with parenthesis) and expand.
Like that: 
```
include_proto!(
    {("/helloworld.rs",hello_world),
     ("/routeguide.rs",routeguide),
     ("/blabla.rs",bla_mod_name),
     /*as much as you want*/ }
);
```
It must depends `build.rs` to generate code. About the issue of directory, I checked the cargo env variables list. For `OUT_DIR`, I guess that I can use `CARGO_TARGET_DIR` to replace it. But for absolute path, the thing we want to do is: to acquire `CARGO_MANIFEST_DIR` of another crate in a different crate (proc macro must be in independent crate). There is no any env variables could access other crate.

Emanon42:
@carllerche I committed updates. Any review? 😅

PR #147, opened 2019-03-31
Apply rustfmt

olix0r:


carllerche:
Could you add a rustfmt check to CI? There is an example here: https://github.com/carllerche/slab/blob/master/.travis.yml

PR #150, opened 2019-04-03
Exercise streaming in routeguide example

mzabaluev:
In the example that features a streaming request, it's more useful
to show how the streams can be used.

The code also demonstrates that the client lives as long as any
request streams are open, so it's not necessary to keep the
`Client` instance around.

### DIFF
--- a/tower-grpc-examples/src/routeguide/client.rs
+++ b/tower-grpc-examples/src/routeguide/client.rs
@@ -19,15 +19,17 @@ extern crate serde_json;
 #[macro_use]
 extern crate serde_derive;
 
-use futures::{Future, Poll};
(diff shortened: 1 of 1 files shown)


tower-grpc-examples/src/routeguide/client.rs
```diff
@@ -55,18 +60,31 @@ pub fn main() {
             RouteGuide::new(conn)
         })
         .and_then(|mut client| {
-            client
```
carllerche: You can probably keep this with the additional calls?

mzabaluev: Redone preserving the original `GetFeature` call.

PR #153, opened 2019-04-08
Added custom_metadata interop test

vilunov:
Issue: #27 

Some issues I encountered
- Test server returns no binary metadata. This may be due to the fact that trailing metadata are not fully supported as mentioned in #90 
- No metadata returned for duplex calls at all

### DIFF
--- a/tower-grpc-interop/src/client.rs
+++ b/tower-grpc-interop/src/client.rs
@@ -721,8 +721,8 @@ impl TestClients {
 
         let stream = stream::iter_ok(vec![make_ping_pong_request(0)]);
         let mut req_stream = Request::new(stream);
-        req_unary.metadata_mut().insert(key1, value1.clone());
-        req_unary.metadata_mut().insert_bin(key2, value2.clone());
+        req_stream.metadata_mut().insert(key1, value1.clone());
+        req_stream.metadata_mut().insert_bin(key2, value2.clone());
 
         let value1_cloned = value1.clone();
         let value2_cloned = value2.clone();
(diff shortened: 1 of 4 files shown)


tower-grpc-interop/src/client.rs
```diff
@@ -700,6 +700,78 @@ impl TestClients {
...
+            .test_client
+            .unary_call(req_unary)
+            .map(move |e| {
+                vec![
+                    test_assert!(
+                        "metadata string must match in unary",
+                        e.metadata().get(key1) == Some(&value1_cloned),
+                        format!("result={:?}", e.metadata().get(key1))
+                    ),
+                    // TODO trailing metadata not implemented?
```
per-gron: I'm a little bit confused about what you mean with this comment. Is this a TODO that states that this is broken and that it should be fixed (implying that the test is not fully completed yet)? If so I think it would be nice to say that explicitly, and perhaps point to issue #27 to provide a little bit of context.

If it is broken, is it broken when run against the Go interop server (implying client side issue) or only when run against the tower-grpc server?

vilunov:
@per-gron Hi, and thanks for the review. By the comment I meant that the test fails when it is run against the Go interop server. I don't know of any mistakes in test implementation, so I believe that it is a client implementation fault, i.e. binary trailing metadata is not read properly, and your #102 seems to support that.

per-gron:
Thanks!

per-gron:
@hawkw do you also want take a look? If not I think this is ready to merge.

carllerche:
Thanks all 👍

PR #154, opened 2019-04-10
Use tower-request-modifier

GeorgeHahn:
Switch from `tower-add-origin` to `tower-request-modifier`. I was hoping this would serve as a clean minimal representation of a trait issue I'm having elsewhere, but it ended up being an easy conversion.

### DIFF
--- a/tower-grpc-interop/src/client.rs
+++ b/tower-grpc-interop/src/client.rs
@@ -10,9 +10,9 @@ extern crate log;
 extern crate prost;
 extern crate rustls;
 extern crate tokio_core;
-extern crate tower_add_origin;
 extern crate tower_grpc;
 extern crate tower_h2;
+extern crate tower_request_modifier;
 
 use std::error::Error;
 use std::fmt;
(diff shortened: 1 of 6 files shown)


seanmonstar:
Restarting CI since tower-http was updated to not need `Default`.

tower-grpc-interop/src/client.rs
```diff
@@ -228,22 +228,24 @@ fn assert_success(
 
 struct TestClients {
     test_client: TestService<
-        tower_add_origin::AddOrigin<
+        tower_request_modifier::RequestModifier<
```
seanmonstar: To save ourselves from this fragility, we should probably change this to a generic (or `BoxService`), but not a blocker.

GeorgeHahn: This is actually something I'm currently struggling with in an internal project - finding a good way to generically pass a (useful) client around. A primary issue I'm seeing is that when passing a client to a function, the function's client lacks all of the client's function implementations due to missing generic bounds.

A potential solution I've been thinking about is implementing a generated client trait. I have not sketched that out yet, but I suspect it could offer a straightforward way to pass a client around without generic bounds-boilerplate(+pain) and other surprises.

seanmonstar: Well, it should hopefully be fine to just be generic over `Service`:

```rust
struct TestClients<T, U>
where
    T: Service<FillInReq, Response = AndResponseTypes>,
    U: Service<FillInReq, Response = AndResponseTypes>,
{
    test_client: TestService<T>,
    unimplemented_client: UnimplementedService<U>,
}
```

GeorgeHahn: I don't fully follow - would you create a separate service for each kind of function call the client can make?

For example, our gRPC Connection struct/client generates something like this:

```
// Multiple unary fn's, each with a different request & response type
pub fn do<R>(&mut self, request: grpc::Request<super::super::do::DoRequest>) -> grpc::unary::ResponseFuture<super::super::print::DoResponse, T::Future, T::ResponseBody>
where T: grpc::GrpcService<R>,
        grpc::unary::Once<super::super::do::DoRequest>: grpc::Encodable<R>,
{ ... }

pub fn a_stream<R>(&mut self, request: grpc::Request<AStreamRequest>) -> grpc::server_streaming::ResponseFuture<StreamItem, T::Future>
where T: grpc::GrpcService<R>,
        grpc::unary::Once<AStreamRequest>: grpc::Encodable<R>,
{ ... }
```

PR #157, opened 2019-04-12
Add tower-hyper support and examples

LucioFranco:
Now that https://github.com/tower-rs/tower-hyper/pull/20 has been merged we can finally support tower-hyper here.

- Add `tower-hyper` feature to main crate and build.
- Add a `tower_hyper::Body` impl for servers.
- Update all examples to use `tower-hyper`


LucioFranco:
@seanmonstar I think it may make sense to have neither? I'm not sure, if not it probably makes sense to support just `tower-hyper` as the default 🤷‍♂️ .

seanmonstar:
Yea, It may be best to disable both by default, and just include a comment in the example files that they are taking advantage of the `tower-hyper` feature.

LucioFranco:
@seanmonstar so it looks like to make `cargo build --all` work we need both features as defaults in tower-grpc and we can make them both optional in build.

tower-grpc-examples/src/helloworld/server.rs
```diff
@@ -42,8 +41,8 @@ pub fn main() {
 
     let new_service = server::GreeterServer::new(Greet);
 
-    let h2_settings = Default::default();
-    let mut h2 = Server::new(new_service, h2_settings, DefaultExecutor::current());
+    let mut server = Server::new(new_service);
+    let http = Http::new().http2_only(true).clone();
```
seanmonstar: Is this clone required?

tower-grpc-examples/src/helloworld/server.rs
```diff
@@ -55,7 +54,7 @@ pub fn main() {
                 return Err(e);
             }
 
-            let serve = h2.serve(sock);
+            let serve = server.serve_with(sock, http.clone());
```
seanmonstar: Does tower-hyper have a way to use `hyper::Server` here, so we can remove the bind and incoming boilerplate?

LucioFranco: I don't think so as there were some compat issues around makeservices and service and hypers service stuff. Though I didnt dive too deep into this.

LucioFranco: We could probably just back this specific functionality into tower-hyper though?

tower-grpc-examples/src/helloworld/server.rs
LucioFranco: I couldn't figure out how to get a non &mut ref since http2_only returns a &mut self. Clone worked do go from a &self to a Self. Any pointers around this would be helpful.

seanmonstar: Oh yea...

```rust
let mut http = Http::new();
http.http2_only(true);
```

tower-grpc-examples/src/helloworld/server.rs
seanmonstar: Yea thats what I meant, the `tower_hyper::Server` type.

tower-grpc-examples/src/helloworld/server.rs
LucioFranco: bleh, can we just add some sort of `build()` to those builder types?

tower-grpc-examples/src/helloworld/server.rs
LucioFranco: Yup it follows tower_h2 in this sense.

seanmonstar:
Oh OK, the failures are because besides the interop tests, we also have a set of "build" tests: https://github.com/tower-rs/tower-grpc/tree/master/tests

LucioFranco:
@seanmonstar right, any idea how to solve that?

PR #173, opened 2019-05-08
Changed HttpBody::Item to Data and poll_buf to poll_data

traxys:
The crate was not compiling, when looking why I found that `HttpBody::Item` had changed name, as did `poll_buf` in the http-body crate, so I changed all occurences

### DIFF
--- a/tower-grpc-build/src/server.rs
+++ b/tower-grpc-build/src/server.rs
@@ -434,11 +434,11 @@ macro_rules! try_ready {
                 .target_generic("T")
                 .impl_trait("tower::HttpBody")
                 .bound("T", &service.name)
-                .associate_type("Item", "<grpc::BoxBody as grpc::Body>::Item")
+                .associate_type("Data", "<grpc::BoxBody as grpc::Body>::Data")
                 .associate_type("Error", "grpc::Status");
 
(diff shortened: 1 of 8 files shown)


traxys:
The same problem is on tower-h2, I could do the same there too

traxys:
There must have been a change in tower-hype `client::Connect::new` because it now takes a single argument, but I do not understand the code to see what should change, so the build still does not work but not because of the Item <=> Data change

seanmonstar:
It *looks* to me that the change is just `Connect::with_builder(connector, settings)`.

LucioFranco:
if I see this correctly this PR can be closed because #174 seems to fix the same things?

traxys:
Yes, I did a second pullrequest with a non clean repo my bad

On 9 May 2019 17:43:42 CEST, Lucio Franco <notifications@github.com> wrote:
>if I see this correctly this PR can be closed because #174 seems to fix
>the same things?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/tower-rs/tower-grpc/pull/173#issuecomment-490957847

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #174, opened 2019-05-09
Fix client with builder

traxys:
There was a renaming of function in tower_hyper I think, so i changed them here too (I am not very good with github and I think i have left the commits of the other pull request I have made, sorry if it poses some problems)


seanmonstar:
Thanks so much!

PR #177, opened 2019-05-28
Remove tower-h2 in favor of tower-hyper in tests

LucioFranco:
This removes all traces of tower-h2 from all the crates. This should get us one step closer to a release!


carllerche:
Any idea why CI isn't passing?

carllerche:
How does tower-hyper know to use H2? Before, H2 happened with "prior knowledge", but my guess is tower-hyper doesn't know out of the box how to do it.

LucioFranco:
@carllerche not too sure why we keep getting a connection refused and that one test keeps failing?

PR #178, opened 2019-05-28
Remove tower-http infavor of http-body

LucioFranco:
This also brings us a step closer to releasing by removing another git dep for its published version.


carllerche:
Thanks 👍

PR #189, opened 2019-07-05
Show usage of cargo:rerun-if-changed in examples

mzabaluev:
As the examples are going to be cribbed by application developers, it's important to demonstrate how rebuilds using tower-grpc-build can be reduced by printing `cargo:rerun-if-changed` instructions for cargo to only rerun the build script if the .proto files change.

### DIFF
--- a/tower-grpc-examples/build.rs
+++ b/tower-grpc-examples/build.rs
@@ -8,6 +8,7 @@ fn main() {
             &["proto/helloworld"],
         )
         .unwrap_or_else(|e| panic!("protobuf compilation failed: {}", e));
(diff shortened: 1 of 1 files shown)


nevi-me:
Hi @mzabaluev, how does this work if one's building multiple proto files in the same build instance? Something like `build(&["proto/one.proto", "proto/two.proto"], &["proto"])`

mzabaluev:
@nevi-me Each file needs its own line, as per [the Cargo book](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script).

Issue #136, opened 2019-03-10
Is there a way to add error details in Status?


csharad:
When returning an error message via `Status`, we can only provide `Code` and message. There is currently no method provided to add details to it as far as I know.

It would be awesome to provide additional metadata to provide the client with additional contexts of the error.

per-gron:
Did you see the `details` field of `Status`? It is for that. The convention is to put a protobuf message in it. I'm not sure if grpc offers a standard proto message type for it though.

csharad:
@per-gron There is a field `details` in `Status` but no way to modify that.

seanmonstar:
It's true there isn't currently a way to add to the `details` method. I'm not especially familiar with the `grpc.Status` message type, but other grpc libraries seem to offer encoding one in the details of a `Status`. That's what this field is supposedly for... We haven't figured out the right way to add them yet, it seems, but I don't really know what that would be, either.

PR #191, opened 2019-07-09
Implement function for creating errors with binary `details` field

GeorgeHahn:
This PR adds a function to create `Status` errors with the `details` field set.

This is not a complete solution - other libraries offer ways to encode a protobuf into this field. This change will allow for some experimentation and manual implementation for error details.

Related: https://github.com/tower-rs/tower-grpc/issues/136

### DIFF
--- a/tower-grpc/src/status.rs
+++ b/tower-grpc/src/status.rs
@@ -59,6 +59,15 @@ impl Status {
         }
     }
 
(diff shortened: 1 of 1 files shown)


GeorgeHahn:
Correct; my hope was to at least make `details` user accessible.

I just pushed a commit that adds the ability to populate details with an encoded protobuf. This is far more useful and may eliminate any desire for `with_raw_details`.

PR #206, opened 2019-09-25
Emit grpc-status in initial response headers on error

### LINKED ISSUE #202: grpc-status never emitted in initial response headers
I modified the `helloworld` server to always return errors:

```rust

impl server::Greeter for Greet {
    type SayHelloFuture = future::FutureResult<Response<HelloReply>, grpc::Status>;

    fn say_hello(&mut self, request: Request<HelloRequest>) -> Self::SayHelloFuture {
        println!("REQUEST = {:?}", request);

        future::err(grpc::Status::new(
            grpc::Code::Unimplemented,
            "I SAID ADIEU!",
        ))
    }
}
```

I'm very surprised to see that this results in TWO response HEADERS frames being emitted:

```
HyperText Transfer Protocol 2                                                                                                                                                                              
    Stream: HEADERS, Stream ID: 1, Length 43, 200 OK                                                                                                                                                       
        Length: 43                                                                                                                                                                                         
        Type: HEADERS (1)                                                                                                                                                                                  
        Flags: 0x04                                                                                                                                                                                        
            .... ...0 = End Stream: False                                                                                                                                                                  
            .... .1.. = End Headers: True                                                                                                                                                                  
            .... 0... = Padded: False                                                                                                                                                                      
            ..0. .... = Priority: False                                                                                                                                                                    
            00.0 ..0. = Unused: 0x00                                                                                                                                                                       
        0... .... .... .... .... .... .... .... = Reserved: 0x0                                                                                                                                            
        .000 0000 0000 0000 0000 0000 0000 0001 = Stream Identifier: 1                                                                                                                                     
        [Pad Length: 0]                                                                                                                                                                                    
        Header Block Fragment: 885f901d75d0620d263d4c4d6564ff75d8749f6196d07abe...                                                                                                                         
        [Header Length: 101]                                                                                                                                                                               
        [Header Count: 3]                                                                                                                                                                                  
        Header: :status: 200 OK                                                                                                                                                                            
            Name Length: 7                                                                                                                                                                                 
            Name: :status                                                                                                                                                                                  
            Value Length: 3                                                                                                                                                                                
            Value: 200                                                                                                                                                                                     
            :status: 200                                                                                                                                                                                   
            Representation: Indexed Header Field                                                                                                                                                           
            Index: 8                                                                                                                                                                                       
        Header: content-type: application/grpc+proto
            Name Length: 12
            Name: content-type
            Value Length: 22
            Value: application/grpc+proto
            content-type: application/grpc+proto
            Representation: Literal Header Field with Incremental Indexing - Indexed Name
            Index: 31
        Header: date: Mon, 02 Sep 2019 20:28:37 GMT
            Name Length: 4
            Name: date
            Value Length: 29
            Value: Mon, 02 Sep 2019 20:28:37 GMT
            date: Mon, 02 Sep 2019 20:28:37 GMT
            Representation: Literal Header Field with Incremental Indexing - Indexed Name
            Index: 33
    Stream: HEADERS, Stream ID: 1, Length 39
        Length: 39
        Type: HEADERS (1)
        Flags: 0x05
            .... ...1 = End Stream: True
            .... .1.. = End Headers: True
            .... 0... = Padded: False
            ..0. .... = Priority: False
            00.0 ..0. = Unused: 0x00
        0... .... .... .... .... .... .... .... = Reserved: 0x0
        .000 0000 0000 0000 0000 0000 0000 0001 = Stream Identifier: 1
        [Pad Length: 0]
        Header Block Fragment: 40889acac8b21234da8f8208bf40899acac8b5254207317f...
        [Header Length: 58]
        [Header Count: 2]
        Header: grpc-status: 12
            Name Length: 11
            Name: grpc-status
            Value Length: 2
            Value: 12
            Representation: Literal Header Field with Incremental Indexing - New Name
        Header: grpc-message: I%20SAID%20ADIEU!
            Name Length: 12
            Name: grpc-message
            Value Length: 17
            Value: I%20SAID%20ADIEU!
            Representation: Literal Header Field with Incremental Indexing - New Name
```

I would expect the `grpc-status` and `grpc-message` headers to be emitted on a single HEADERS frame, and not to be manifested in a trailers frame.  (It [_seems_ like grpc-go does this](https://github.com/grpc/grpc-go/blob/4d39b489547474addb4cf6a2368c1ea05450c11e/server.go#L995-L1007) as I describe, though I need to confirm in a test...).

### the real problem...

The problem I'm trying to solve, however, is the following: Because tower-grpc always emits `grpc-status` headers in trailers (and never in headers), clients cannot determine when streaming-response methods fail early!

For example, let's modify the helloword's Greeter service with a streaming-response method like:
```protobuf
service Greeter {
  rpc StreamHello (HelloRequest) returns (stream HelloReply) {}
}
```

and we provide an implementation like:
```rust
    type StreamHelloStream = stream::Empty<HelloReply, grpc::Status>;
    type StreamHelloFuture = future::FutureResult<Response<Self::StreamHelloStream>, grpc::Status>;

    fn stream_hello(&mut self, request: Request<HelloRequest>) -> Self::StreamHelloFuture {
        println!("REQUEST = {:?}", request);

        future::err(grpc::Status::new(
            grpc::Code::Unimplemented,
            "I SAID ADIEU!",
        ))
    }
```

A tower-grpc _client_ for this service will never be able to fail the response future from `stream_hello`! Specifically, we can never enter this code to cause the response future to fail:
https://github.com/tower-rs/tower-grpc/blob/ea4d6fd95563ae54d0ec900ed01ce0e44b46e404/tower-grpc/src/client/streaming.rs#L49-L56

## conclusion

The tower-grpc server should be modified to encode grpc headers in the initial response headers when the status is a failure and no message is encoded.

kleimkuhler:
### Summary

Clients are unable to determine when streaming response methods fail early
because `grpc-status` and `grpc-message` headers are not included in the intial
HEADERS frame. Failures in response methods result in these headers being
present in trailers.

### Solution

`tower_grpc::Codec` now dispatches to its inner
`tower_grpc::generic::Encode<Encode<_, _>>::is_end_stream` method to determine
if the EOS flag should be set on the encoded gRPC response body.

A new `Empty` variant has been added to `Encode::EncodeInner`. When an error
occurs polling for a response, an HTTP response is created with an empty body.
The error status is now added to the headers of the response, and EOS flag is
set by matching on `Empty`.

As @olix0r demonstrated in #202, a server that responds with an error results in
two HEADERS frames. It now results in one as shown below.

```
HyperText Transfer Protocol 2
    Stream: HEADERS, Stream ID: 1, Length 94, 200 OK
        Length: 94
        Type: HEADERS (1)
        Flags: 0x05
            .... ...1 = End Stream: True
            .... .1.. = End Headers: True
            .... 0... = Padded: False
            ..0. .... = Priority: False
            00.0 ..0. = Unused: 0x00
        0... .... .... .... .... .... .... .... = Reserved: 0x0
        .000 0000 0000 0000 0000 0000 0000 0001 = Stream Identifier: 1
        [Pad Length: 0]
        Header Block Fragment: 885f901d75d0620d263d4c4d6564ff75d8749f40889acac8...
        [Header Length: 174]
        [Header Count: 5]
        Header: :status: 200 OK
            Name Length: 7
            Name: :status
            Value Length: 3
            Value: 200
            :status: 200
            Representation: Indexed Header Field
            Index: 8
        Header: content-type: application/grpc+proto
            Name Length: 12
            Name: content-type
            Value Length: 22
            Value: application/grpc+proto
            content-type: application/grpc+proto
            Representation: Literal Header Field with Incremental Indexing - Indexed Name
            Index: 31
        Header: grpc-status: 12
            Name Length: 11
            Name: grpc-status
            Value Length: 2
            Value: 12
            Representation: Literal Header Field with Incremental Indexing - New Name
        Header: grpc-message: I%20SAID%20ONE%20FRAME%20PLEASE!
            Name Length: 12
            Name: grpc-message
            Value Length: 32
            Value: I%20SAID%20ONE%20FRAME%20PLEASE!
            Representation: Literal Header Field with Incremental Indexing - New Name
        Header: date: Wed, 25 Sep 2019 03:54:06 GMT
            Name Length: 4
            Name: date
            Value Length: 29
            Value: Wed, 25 Sep 2019 03:54:06 GMT
            date: Wed, 25 Sep 2019 03:54:06 GMT
            Representation: Literal Header Field with Incremental Indexing - Indexed Name
            Index: 33
```

Closes #202

Signed-off-by: Kevin Leimkuhler <kleimkuhler@icloud.com>

### DIFF
--- a/tower-grpc/src/body.rs
+++ b/tower-grpc/src/body.rs
@@ -60,12 +60,6 @@ pub struct BoxBody {
 
 struct MapBody<B>(B);
 
-// #[derive(Debug)]
-// pub struct NoBody {}
-
-// #[derive(Debug)]
-// pub struct NoData;
-
 // ===== impl BoxBody =====
 
 impl BoxBody {
@@ -130,56 +124,6 @@ where
         self.0.poll_trailers().map_err(Status::map_error)
     }
 }
-
-// ===== impl NoBody =====
-
-// impl NoBody {
-//     pub fn new() -> Self {
-//         NoBody {}
-//     }
-// }
-
-// impl Body for NoBody {
-//     type Data = NoData;
-//     type Error = Error;
-
-//     fn is_end_stream(&self) -> bool {
-//         true
-//     }
-
-//     fn poll_data(&mut self) -> Poll<Option<Self::Data>, Self::Error> {
-//         Ok(None.into())
-//     }
-
-//     fn poll_trailers(&mut self) -> Poll<Option<http::HeaderMap>, Self::Error> {
-//         Ok(None.into())
-//     }
-// }
-
-// impl Stream for NoBody {
-//     type Item = ();
-//     type Error = Never;
-
-//     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
-//         Ok(Async::Ready(None))
-//     }
-// }
-
-// impl Sealed for NoBody {}
-
-// ===== impl NoData =====
-
-// impl Buf for NoData {
-//     fn remaining(&self) -> usize {
-//         0
-//     }
-
-//     fn bytes(&self) -> &[u8] {
-//         &[]
-//     }
-
-//     fn advance(&mut self, _cnt: usize) {}
-// }
 
 mod sealed {
     pub trait Sealed {}
--- a/tower-grpc/src/body.rs
+++ b/tower-grpc/src/body.rs
@@ -60,6 +60,11 @@ pub struct BoxBody {
 
 struct MapBody<B>(B);
 
+#[derive(Debug, Default)]
+struct EmptyBody {
+    _p: (),
+}
+
 // ===== impl BoxBody =====
 
 impl BoxBody {
@@ -75,6 +80,13 @@ impl BoxBody {
         B::Data: Into<Bytes>,
     {
         BoxBody::new(Box::new(MapBody(inner)))
+    }
+
+    /// Create a new `BoxBody` that is empty.
+    pub fn empty() -> Self {
+        Self {
+            inner: Box::new(EmptyBody::default()),
+        }
     }
 }
 
@@ -122,6 +134,23 @@ where
 
     fn poll_trailers(&mut self) -> Poll<Option<http::HeaderMap>, Self::Error> {
         self.0.poll_trailers().map_err(Status::map_error)
+    }
+}
+
+impl HttpBody for EmptyBody {
+    type Data = BytesBuf;
+    type Error = Status;
+
+    fn is_end_stream(&self) -> bool {
+        true
+    }
+
+    fn poll_data(&mut self) -> Poll<Option<Self::Data>, Self::Error> {
+        Ok(None.into())
+    }
+
+    fn poll_trailers(&mut self) -> Poll<Option<http::HeaderMap>, Self::Error> {
+        Ok(None.into())
     }
 }
 
--- a/tower-grpc/src/body.rs
+++ b/tower-grpc/src/body.rs
@@ -1,11 +1,12 @@
 use self::sealed::Sealed;
-use crate::error::Error;
+use crate::error::{Error, Never};
 use crate::Status;
 
 use bytes::{Buf, Bytes, IntoBuf};
-use futures::{try_ready, Poll};
+use futures::{try_ready, Async, Poll, Stream};
 pub use http_body::Body as HttpBody;
 use std::fmt;
+use std::marker::PhantomData;
 
 type BytesBuf = <Bytes as IntoBuf>::Buf;
 
@@ -59,6 +60,14 @@ pub struct BoxBody {
 }
 
 struct MapBody<B>(B);
+
+#[derive(Debug)]
+pub struct NoBody<T> {
+    pub(crate) _marker: PhantomData<T>,
+}
+
+#[derive(Debug)]
+pub struct NoData;
 
 // ===== impl BoxBody =====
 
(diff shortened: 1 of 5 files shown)


tower-grpc/src/generic/codec.rs
```diff
@@ -81,7 +81,7 @@ enum EncodeInner<T, U> {
         /// The source of messages to encode
         inner: U,
     },
-    Err(Status),
+    Err,
```
LucioFranco: Does this now remove the ability to return a trailing status code when the inner stream bails out?

tower-grpc/src/generic/codec.rs
```diff
@@ -189,6 +189,9 @@ where
     type Error = Status;
 
     fn is_end_stream(&self) -> bool {
+        if let EncodeInner::Err = self.inner {
```
LucioFranco: minor nit: can we drop the return and use an else here?

LucioFranco:
So I'm not sure if this is correct, I took a bit of a different solution to this. It was to encode an empty body and totally avoid using the Encode struct if we got an inital error status. This should then allow the error status to be sent in one header that half closes.

tower-grpc/src/generic/codec.rs
kleimkuhler: With this change, if an error occurs on the inner stream `status` is now included in the headers and `Encode::is_end_stream` would return `true`.

If the inner stream does not initially error, but bails out a later point, would it be unable to send the `status` back to the client because it is now part of the headers? If that is the change in behavior that was not intended!

This starts to get into the strategy used to solve this, so I'll respond to your comment down below.

kleimkuhler:
@LucioFranco So I had in initially considered making `ResponseFuture<_, _>::Item = http::Response<Option<Encode<_, _>>>` This way, when an error occurs we can return no `Encode` struct and handle the error separately.

Would this still be an issue though with what you pointed out about when the inner stream bails out?

LucioFranco:
@kleimkuhler so I think an issue with the `Option` model is that `None` doesn't implement `Body`? I think we can actually fix this bug but not even touching encoding at all but when the inital Error happens we just supply an empty body and set the header. I found this fixed a few interop tests with grpc-go.

kleimkuhler:
@LucioFranco Thanks for the review on this! I've updated the changes to return an empty response with gRPC codes added to the headers without removing the `status` field from `EncodeInner::Err`.

The above example still works as would be expected--one frame is returned.

tower-grpc/src/body.rs
```diff
@@ -60,6 +60,11 @@ pub struct BoxBody {
 
 struct MapBody<B>(B);
 
+#[derive(Debug, Default)]
+struct EmptyBody {
```
LucioFranco: I don't think this is being used anywhere is it?

kleimkuhler: Right it is no longer used; it's been removed!

LucioFranco (approved):
LGTM!

tower-grpc/src/generic/server/streaming.rs
```diff
@@ -44,12 +44,14 @@ where
             Ok(Async::NotReady) => return Ok(Async::NotReady),
             Err(status) => {
                 // Construct http response
-                let mut response = Response::new(Encode::error(status)).into_http();
-                // Set the content type
+                let mut response = Response::new(Encode::empty()).into_http();
+
+                // Set the content type and add the gRPC error codes
                 response.headers_mut().insert(
```
olix0r: Excuse the late question but I wonder if we should explicitly encode `Content-length: 0`?  Or can we trust something else to do that?

LucioFranco: @seanmonstar ^ thoughts?

LucioFranco: I don't see anything explicitly within the gRPC spec for this so not sure if this is right or what hyper might add.

PR #208, opened 2019-09-25
UPDATE: Update the dependency(branch) of the h2 and typer

UkonnRa:
However, this branch is still not working, for:
```
107 | impl<T: Unpin> TryStream for Once<T> {
    |                ^^^^^^^^^ the trait `futures_core::stream::Stream` is not implemented for `generic::server::unary::Once<T>`
    |
    = note: required because of the requirements on the impl of `futures_core::stream::private_try_stream::Sealed` for `generic::server::unary::Once<T>`
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+.idea
--- a/tower-grpc-examples/Cargo.toml
+++ b/tower-grpc-examples/Cargo.toml
@@ -5,9 +5,9 @@ edition = "2018"
 authors = ["Carl Lerche <me@carllerche.com>"]
 publish = false
 
-# [[bin]]
-# name = "helloworld-server"
-# path = "src/helloworld/server.rs"
+ [[bin]]
+ name = "helloworld-server"
+ path = "src/helloworld/server.rs"
 
 [[bin]]
 name = "helloworld-client"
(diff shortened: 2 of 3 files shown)


LucioFranco:
@UkonnRa the `std-future` branch is currently very far behind and in a stale state right now. I would recommend waiting a week or so and well have something that will make this all much easier to interact with std::future::Future.

UkonnRa:
So could I access the latest branch of `tower-rs` with the latest `rs-futures 0.3`?

LucioFranco:
@UkonnRa so `tower` has a `0.3.x` branch for `0.3` futures. `tower-grpc` currently has not been updated as we have focused on updating the rest of the stack. But we will have something very shortly for this. So I would wait to use tower-grpc and std future for now. The `std-future` branch here was a test to see if we could upgrade tower-grpc and we ran into a few issues. But it is our top priority right now.

UkonnRa:
OK I get it, so `tower-grpc` is blocked by `tower`, `tower-grpc` will be updated after the `tower` side work is finished. That's fair~

LucioFranco:
@UkonnRa I will give an update to this repo, probably the readme and an issue once we have the new stuff out. Thanks for the patience <3

